PR #29, opened 2017-02-06
Update instructions for better eslint support

iatsiuk:
Without this steps linting not working


octref:
Thanks for the PR! However it seems to me more people are using https://github.com/vuejs/eslint-plugin-vue, and it's under the official vuejs organization.

So I'd say either leave the setup up to users, or recommend `vuejs/eslint-plugin-vue`.

iatsiuk:
ok, you're right

PR #94, opened 2017-03-16
Add vue.d.ts and typescript support

### LINKED ISSUE #75: Better intellisense for `this` references in script block
I'm working on a [plugin for the typescript language service](https://github.com/sandersn/vue-ts-plugin) that will provide better intellisense for `this` references in an options object inside the script block of a vue file. In a .js (or .ts) file, the call to `new Vue(...)` can provide a contextual type to the object literal.

```ts
import Vue from 'vue'
const app = new Vue({
  data: { greeting: 'hi' },
  methods: {
    m1() {
      // Typescript language service can now offer 'greeting' as a completion
      // and also knows that greeting: string.
      // (in both .js and .ts files)
      this./**/greeting
    }
  }
});
```

But inside a Vue file, the typical code is just `export default { ... }`, so there is no such clue to the language service that the default-exported object literal is special.

```vue
<script>
export default {
  data: { greeting: 'hi' },
  methods: {
    m1() {
      // no completions here, and greeting: any
      this./**/greeting
    }
  }
}
```

To fix this, the plugin generates a synthetic `new Vue(...)` call to wrap the object literal, plus a synthetic `import Vue from "vue"` statement.

Right now the installation process is complicated, but I hope that it will be as easy as making sure that vetur is installed, then:

```sh
$ npm install vue-ts-plugin
```

Add a tsconfig.json with a plugins entry:

```json
{
    "compilerOptions": {
        "plugins": [{ "name": "sample-ts-plugin" }]
    },
    "files": [
        "test.vue",
        "other.vue",
    ]
}
```

However, it does not currently work so I'm investigating to see what changes are needed. The plugin loads and starts logging but the types and completions are still the old ones. I'll create PRs if there are changes needed in vetur.

A few notes:

1. The plugin depends on unfinished code in Typescript that may be release by 2.3.
2. vue's d.ts file will need to be updated after Typescript 2.3's release to use the new `ThisType` feature.
3. @octref mentions that vuex also doesn't work well right now. I need to learn more about vuex, but it looks like the `ThisType` feature would also address the basic vuex transformation.

sandersn:
1. Add an implicit `import Vue from 'vue'` and `new Vue(...)` around the default exported object literal. This provides a lot better intellisense than just a bare object literal because the Typescript language server knows that the object will be used by Vue.
2. Resolve other .vue files so that `import other from './other.vue'` provides correct info.
3. If lang='typescript' is specified, put the TS language service in TS mode: all errors are reported and completions include only symbols that are known to be correct.
4. <strike>Also fix a bug in htmlServerMain so that errors are actually reported.</strike> Looks like this is already fixed in master.

Fixes #75

### DIFF
--- a/server/src/modes/javascriptMode.ts
+++ b/server/src/modes/javascriptMode.ts
@@ -3,8 +3,8 @@ import { SymbolInformation, SymbolKind, CompletionItem, Location, SignatureHelp,
 import { LanguageMode } from './languageModes';
 import { getWordAtText, isWhitespaceOnly, repeat } from '../utils/strings';
 import { HTMLDocumentRegions } from './embeddedSupport';
+import Uri from 'vscode-uri';
 import path = require('path');
-import url = require('url');
 
 import { createUpdater, parseVue, isVue } from './typescriptMode';
 
@@ -36,7 +36,7 @@ export function getJavascriptMode(documentRegions: LanguageModelCache<HTMLDocume
   function updateCurrentTextDocument(doc: TextDocument) {
     if (!currentTextDocument || doc.uri !== currentTextDocument.uri || doc.version !== currentTextDocument.version) {
       currentTextDocument = jsDocuments.get(doc);
-      const fileName = trimFileUri(currentTextDocument.uri);
+      const fileName = trimFileUri(doc.uri);
       if (docs.has(fileName) && currentTextDocument.languageId !== docs.get(fileName).languageId) {
         // if languageId changed, we must restart the language service; it can't handle file type changes
         compilerOptions.allowJs = docs.get(fileName).languageId !== 'typescript';
@@ -69,11 +69,16 @@ export function getJavascriptMode(documentRegions: LanguageModelCache<HTMLDocume
   let host: ts.LanguageServiceHost = {
     getCompilationSettings: () => compilerOptions,
     getScriptFileNames: () => files,
-    getScriptVersion: filename => versions.has(filename) ? versions.get(filename).toString() : '0',
+    getScriptVersion(filename) { 
+      filename = normalizeFileName(filename);
+      return versions.has(filename) ? versions.get(filename).toString() : '0';
+    },
     getScriptKind(fileName) { 
       if(isVue(fileName)) {
+        const uri = Uri.file(fileName);
+        fileName = uri.fsPath;
         const doc = docs.get(fileName) ||
-          jsDocuments.get(TextDocument.create('file://' + fileName, 'vue', 0, ts.sys.readFile(fileName)));
+          jsDocuments.get(TextDocument.create(uri.toString(), 'vue', 0, ts.sys.readFile(fileName)));
         return doc.languageId === 'typescript' ? ts.ScriptKind.TS : ts.ScriptKind.JS;
       }
       else {
@@ -89,9 +94,10 @@ export function getJavascriptMode(documentRegions: LanguageModelCache<HTMLDocume
           return ts.resolveModuleName(name, containingFile, compilerOptions, ts.sys).resolvedModule;
         }
         else {
-          const resolvedFileName = path.join(path.dirname(containingFile), name);
+          const uri = Uri.file(path.join(path.dirname(containingFile), name));
+          const resolvedFileName = uri.fsPath;
           const doc = docs.get(resolvedFileName) ||
-            jsDocuments.get(TextDocument.create('file://' + name, 'vue', 0, ts.sys.readFile(resolvedFileName)));
+            jsDocuments.get(TextDocument.create(uri.toString(), 'vue', 0, ts.sys.readFile(resolvedFileName)));
           return {
             resolvedFileName,
             extension: doc.languageId === 'typescript' ? ts.Extension.Ts : ts.Extension.Js,
@@ -100,6 +106,7 @@ export function getJavascriptMode(documentRegions: LanguageModelCache<HTMLDocume
       });
     },
     getScriptSnapshot: (fileName: string) => {
+      fileName = normalizeFileName(fileName);
       let text = docs.has(fileName) ? docs.get(fileName).getText() : (ts.sys.readFile(fileName) || '');
       if (isVue(fileName)) {
         // Note: This is required in addition to the parsing in embeddedSupport because
@@ -354,7 +361,11 @@ export function getJavascriptMode(documentRegions: LanguageModelCache<HTMLDocume
 };
 
 function trimFileUri(uri: string): string {
-  return url.parse(uri).path;
+  return Uri.parse(uri).fsPath;
+}
+
+function normalizeFileName(fileName: string): string {
+  return Uri.file(fileName).fsPath;
 }
 
 function convertRange(document: TextDocument, span: { start: number, length: number }): Range {
--- a/server/src/modes/javascriptMode.ts
+++ b/server/src/modes/javascriptMode.ts
@@ -8,7 +8,11 @@ import url = require('url');
 
 import { createUpdater, parseVue, isVue } from './typescriptMode';
 
+<<<<<<< HEAD
 import * as ts from 'typescript';
+=======
+const FILE_NAME = 'vscode://javascript/1';  // the same 'file' is used for all contents
+>>>>>>> master
 
 const JS_WORD_REGEX = /(-?\d*\.\d\w*)|([^\`\~\!\@\#\%\^\&\*\(\)\-\=\+\[\{\]\}\\\|\;\:\'\"\,\.\<\>\/\?\s]+)/g;
 
--- a/server/src/modes/javascriptMode.ts
+++ b/server/src/modes/javascriptMode.ts
@@ -8,11 +8,7 @@ import url = require('url');
 
 import { createUpdater, parseVue, isVue } from './typescriptMode';
 
-<<<<<<< HEAD
 import * as ts from 'typescript';
-=======
-const FILE_NAME = 'vscode://javascript/1';  // the same 'file' is used for all contents
->>>>>>> master
 
 const JS_WORD_REGEX = /(-?\d*\.\d\w*)|([^\`\~\!\@\#\%\^\&\*\(\)\-\=\+\[\{\]\}\\\|\;\:\'\"\,\.\<\>\/\?\s]+)/g;
 
--- a/server/src/modes/javascriptMode.ts
+++ b/server/src/modes/javascriptMode.ts
@@ -69,10 +69,12 @@ export function getJavascriptMode(documentRegions: LanguageModelCache<HTMLDocume
   let host: ts.LanguageServiceHost = {
     getCompilationSettings: () => compilerOptions,
     getScriptFileNames: () => files,
-    getScriptVersion: filename => versions.has(filename) ? versions.get(filename).toString() : '1',
+    getScriptVersion: filename => versions.has(filename) ? versions.get(filename).toString() : '0',
     getScriptKind(fileName) { 
-      if(isVue(fileName) && docs.has(fileName)) {
-        return docs.get(fileName).languageId === 'typescript' ? ts.ScriptKind.TS : ts.ScriptKind.JS;
+      if(isVue(fileName)) {
+        const doc = docs.get(fileName) ||
+          jsDocuments.get(TextDocument.create('file://' + fileName, 'vue', 0, ts.sys.readFile(fileName)));
+        return doc.languageId === 'typescript' ? ts.ScriptKind.TS : ts.ScriptKind.JS;
       }
       else {
         // NOTE: Typescript 2.3 should export getScriptKindFromFileName. Then this cast should be removed.
@@ -82,13 +84,20 @@ export function getJavascriptMode(documentRegions: LanguageModelCache<HTMLDocume
     resolveModuleNames(moduleNames: string[], containingFile: string): ts.ResolvedModule[] {
       // in the normal case, delegate to ts.resolveModuleName
       // in the relative-imported.vue case, manually build a resolved filename
-      return moduleNames.map(name => 
-        path.isAbsolute(name) || !isVue(name) ? 
-          ts.resolveModuleName(name, containingFile, compilerOptions, ts.sys).resolvedModule : 
-          {
-            resolvedFileName: path.join(path.dirname(containingFile), name),
-            extension: docs.has(name) && docs.get(name).languageId === 'typescript' ? ts.Extension.Ts : ts.Extension.Js,
-          })
+      return moduleNames.map(name => {
+        if (path.isAbsolute(name) || !isVue(name)) {
+          return ts.resolveModuleName(name, containingFile, compilerOptions, ts.sys).resolvedModule;
+        }
+        else {
+          const resolvedFileName = path.join(path.dirname(containingFile), name);
+          const doc = docs.get(resolvedFileName) ||
+            jsDocuments.get(TextDocument.create('file://' + name, 'vue', 0, ts.sys.readFile(resolvedFileName)));
+          return {
+            resolvedFileName,
+            extension: doc.languageId === 'typescript' ? ts.Extension.Ts : ts.Extension.Js,
+          };
+        }
+      });
     },
     getScriptSnapshot: (fileName: string) => {
       let text = docs.has(fileName) ? docs.get(fileName).getText() : (ts.sys.readFile(fileName) || '');
--- a/server/src/modes/javascriptMode.ts
+++ b/server/src/modes/javascriptMode.ts
@@ -24,9 +24,10 @@ export function getJavascriptMode(documentRegions: LanguageModelCache<HTMLDocume
   let compilerOptions: ts.CompilerOptions = {
     allowNonTsExtensions: true,
     allowJs: true,
-    lib: ['lib.es6.d.ts'],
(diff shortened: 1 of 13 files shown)


sandersn:
@mhegazy you were interested in this PR, you said.

octref:
Thanks a lot for the PR!

Just trying it out on this repo: https://github.com/vuejs/vue-hackernews-2.0, but IntelliSense didn't work:

- `Vue.` doesn't prompt [`Vue.extend`](https://vuejs.org/v2/api/#Vue-extend) in *.vue files (works in js files). Thought this would work with the implicit addition of `import Vue from 'vue'`
- Newly opened *.vue files don't have IntelliSense and always show "loading"
  image
- As for `new Vue(...)`, if I have `Item.vue` with a prop `count`, and in another vue file I import `Item.vue`, is it that
  - `Item.` is supposed to prompt all values on the `new Vue(...)` object?
  - `Item.props.` should prompt `count`?

Just a rough look at the code, it seems I need a properly configured jsconfig/tsconfig.

Would you mind adding appropriate config for a sample vue project so it would work as you intended? Either https://github.com/vuejs/vue-hackernews-2.0 or something from here https://github.com/vuejs/vue/tree/dev/examples

octref:
Oh and in vetur's language server I'm using TS 2.2.1. Is is I should build from typescript's master and link it locally to vetur's language server?

sandersn:
I'll test with vue-hackernews-2.0, in particular without a tsconfig. I thought the change would work without a tsconfig, but I didn't test that way, so I'm not surprised that it doesn't. 

* I don't think you need any special version of TS. I'll double check though.
* "Loading..." usually means the language service asserted and hit the `debugger` statement we have there. I'll check it out. (This is probably the cause of no completion for `Vue.extends` too.)
* Is this the code you're thinking of?:

```js
// @filename:Item.vue
export default {
  props: ['count'],
  data: {
    d1: 12
  }
  // etc ...
}

// @filename:SomeComponent.vue
//          (or somethingElse.js)
import Item from './Item.vue'
Item./**/ // should see d1, $data, $parent, etc
Item.props./**/ // should see count
```

The short answer is yes, those should work. But `vue.d.ts` doesn't support a lot of these things yet. @DanielRosenwasser is working on improvements to `vue.d.ts` that rely on Typescript 2.3 features. Once 2.3 ships with VS Code, the experience will get a lot better.

Unfortunately, better `this` inference relies on 2.3 as well. Even without 2.3, though, this change makes module resolution work, and will pick up any other improvements to vue.d.ts that happen in the meantime.

octref:
I tested with lodash and @types/lodash, module resolution works great. Thanks!

> Is this the code you're thinking of?

Something like that. I'm wondering if there is a way to get the Vue wrapper object on `Item.vue` at LS side so I can use its info to power html suggestion. `Item./**/` IntelliSense in JS region wouldn't be helpful. But this kind of IntelliSense in html template region would be godly:

```
<i/**/ // Prompt item

<item :/**/> // Prompt count and other props
```

sandersn:
Yes, from typescript's perspective the object from Item.vue is already a Vue object because it was called from `new Vue(...)`. However, you would have to make the typescriptMode.parse code *much* smarter to use that in HTML. Currently it just replaces things outside `<script>` with spaces. If it replaced non-standard HTML elements with valid typescript, which would each have to be smaller than the original HTML so they didn't overflow the original, it could provide completions on that typescript.

I think the Angular guys have done or were thinking about doing something like that for Angular 2. I don't know Angular though, so I'm not sure of the details.

You might also do some other clever thing with a virtual TextDocument for each HTML element, which would make the `host` code in javascriptMode much more complicated. And I don't know whether you could do anything like validation, only completions, because the offsets from the language service would likely be quite wrong.

sandersn:
One update and one question:

1. I confirmed that this works with Typescript on master and 2.2. So both typescript@next and the current Typescript should be fine.
2. Should the inserted call be `Vue.extends` instead of `new Vue`?

octref:
> Should the inserted call be Vue.extends instead of new Vue?

I think `new Vue`.

server/src/modes/typescriptMode.ts
```diff
@@ -0,0 +1,76 @@
...
+            zero(ts.createIdentifier('Vue')),
+            zero(ts.createIdentifier('Vue'))))])))),
+      zero(ts.createLiteral('vue'))));
+    sourceFile.statements.unshift(vueImport);
+
+    // 2. find the export default and wrap it in `new Vue(...)` if it exists and is an object literal
+    //       (the span of the construct call is the same as the object literal)
+    const objectLiteral = (exportDefaultObject as ts.ExportAssignment).expression as ts.ObjectLiteralExpression;
+    const o = <T extends ts.TextRange>(n: T) => ts.setTextRange(n, objectLiteral);
+    const vue = ts.setTextRange(ts.createIdentifier('Vue'), { pos: objectLiteral.pos, end: objectLiteral.pos + 1 });
```
octref: Would you mind giving `o` and `n` more descriptive names? I'm not quite familiar with TS compilers but would want to fiddle with this part of code.

octref:
I'll merge once you fix the one comment.

Meanwhile thanks for giving me a starting point -- I guess for now I'll just use TS to parse the object literal's AST, find out props / methods / data manually to provide some IntelliSense.

image

sandersn:
I improved the naming, structure and documentation of the AST creation utilities. Let me know if you have other questions about creating or altering ASTs.

I'm still investigating the failure to resolve 'vue'. It works fine when I use './node_modules/vue/types/index' instead. My test project may have incorrect settings in tsconfig.

server/src/modes/javascriptMode.ts
```diff
@@ -1,55 +1,97 @@
...
+export function getJavascriptMode(documentRegions: LanguageModelCache<HTMLDocumentRegions>, workspacePath: string): LanguageMode {
+  let jsDocuments = getLanguageModelCache<TextDocument>(10, 60, document => {
+    const vueDocument = documentRegions.get(document);
+    if (vueDocument.getLanguagesInDocument().indexOf('typescript') > -1) {
+      return vueDocument.getEmbeddedDocument('typescript');
+    }
+    return vueDocument.getEmbeddedDocument('javascript');
+  });
 
   let compilerOptions: ts.CompilerOptions = { allowNonTsExtensions: true, allowJs: true, lib: ['lib.es6.d.ts'], target: ts.ScriptTarget.Latest, moduleResolution: ts.ModuleResolutionKind.Classic };
```
mhegazy: For this i would defere to `ts.getDefaultCompilerOptions` and just set:
```ts
let compilerOptions = ts.getDefaultCompilerOptions();
if (docs[fileName].languageId !== 'typescript') {
     copilerOptions.allowJs = true
}
compilerOptions.allowNonTsExtensions = true;
````

File: server/src/modes/javascriptMode.ts [hunk 3]
```diff
@@ -1,55 +1,97 @@
...
+  (ts as any).updateLanguageServiceSourceFile = updateLanguageServiceSourceFile;
+  const configFile = ts.findConfigFile(workspacePath, ts.sys.fileExists, 'tsconfig.json') ||
+    ts.findConfigFile(workspacePath, ts.sys.fileExists, 'jsconfig.json');
+  const files = ts.parseJsonConfigFileContent({},
+    ts.sys,
+    workspacePath,
+    compilerOptions,
+    configFile,
+    undefined,
+    [{ extension: 'vue', isMixedContent: true }]).fileNames;
```
mhegazy: should also set compilerOptions to the options from the file. we also need to set `compilerOptions.allowNonTsExtensions = true;` all the time to allow `.vue` files.

server/src/modes/javascriptMode.ts
mhegazy: I take that back.. 

```ts
{
   "allowJs": docs[fileName].languageId !== 'typescript',
   "lib" : ["dom", "esNext"],
   "target" : ["esNext"],
   "module": ts.ModuleKind.CommonJs
}
```

server/src/modes/typescriptMode.ts
```diff
@@ -0,0 +1,82 @@
...
+  if (output && output.script && output.script.content) {
+    return output.script.content;
+  }
+  else {
+    return text;
+  }
+}
+
+export function createUpdater() {
+  const clssf = ts.createLanguageServiceSourceFile;
```
mhegazy: these should be added to the LanguageServiceHost API in TS 2.3, I would also add a comment here with that to allow removing this global function hijacking.

server/src/modes/javascriptMode.ts
```diff
@@ -1,55 +1,97 @@
...
-        return String(scriptFileVersion);
+    getScriptFileNames: () => files,
+    getScriptVersion: filename => filename in versions ? versions[filename].toString() : '1',
+    getScriptKind(fileName) { 
+      if(isVue(fileName) && docs[fileName]) {
+        return docs[fileName].languageId === 'typescript' ? ts.ScriptKind.TS : ts.ScriptKind.JS;
       }
-      return '1'; // default lib an jquery.d.ts are static
+      else {
+        return (ts as any).getScriptKindFromFileName(fileName);
```
mhegazy: We need to expose this on the API for TS 2.3, also add a comment allowing removal of this cast.

server/src/modes/javascriptMode.ts
```diff
@@ -1,55 +1,97 @@
...
+    if (vueDocument.getLanguagesInDocument().indexOf('typescript') > -1) {
+      return vueDocument.getEmbeddedDocument('typescript');
+    }
+    return vueDocument.getEmbeddedDocument('javascript');
+  });
 
   let compilerOptions: ts.CompilerOptions = { allowNonTsExtensions: true, allowJs: true, lib: ['lib.es6.d.ts'], target: ts.ScriptTarget.Latest, moduleResolution: ts.ModuleResolutionKind.Classic };
   let currentTextDocument: TextDocument;
-  let scriptFileVersion: number = 0;
+  let versions: ts.MapLike<number> = {};
```
mhegazy: you can use ES6 Map

sandersn: done

server/src/modes/javascriptMode.ts
sandersn: added comment. I'll send out a PR for the API change after this.

server/src/modes/typescriptMode.ts
sandersn: added comment where the hijacking happens. This part of the code may remain the same since it's basically the same way that the plugin API allows users to wrap the language service.

server/src/modes/javascriptMode.ts
sandersn: "esnext" didn't ship as a lib target in 2.2, so I'll use "es2017" for now.
(It was committed Dec 30, but I guess it was part of a PR that got merged later.)

server/src/modes/javascriptMode.ts
sandersn: done

sandersn:
All right, I addressed all of @mhegazy's comments. I think it's ready to go in but I'll spend time testing without a tsconfig now to make sure.

sandersn:
The failure to resolve `import Vue from 'vue'` was due to missing `allowSyntheticDefaultImports: true` and having `moduleResolution: ts.ModuleResolutionKind.Classic`. It needs to be `NodeJs`.

sandersn:
OK, vue-hackernews-2.0 is now working and gives nice (?) errors when you switch to `lang="typescript"`. Actually the vue.d.ts needs some updates to make this mode actually nice because it issues some bogus errors. 

For both javascript and typescript, the completions are pretty nice, though.

I'm chasing down some bugs to do with changing the language back and forth between javascript and typescript. I think this is ready to merge as-is though. I'll send another PR if/when I track down the language-changing bug.

sandersn:
Actually, the language-change bug repros when opening or closing files. Probably you should wait to merge this until I track it down.

sandersn:
If you open a .vue file with `lang='typescript'` and then open one without, then the newly opened file gets initially set as typescript, even though it shouldn't. But then when you interact with the language service, it resets the type to javascript, which hits the assert.

octref:
> Actually, the language-change bug repros when opening or closing files. Probably you should wait to merge this until I track it down.

I'll take a closer look at the code tonight. Let me know when you are done.

BTW just checked out the latest one. Works great on `vue-hackernews`.

> gives nice (?) errors... Actually the vue.d.ts needs some updates to make this mode actually nice because it issues some bogus errors.

I suppose other than updating `vue.d.ts`, on vetur side I also need to use this feature introduced with https://github.com/Microsoft/TypeScript/pull/14141

> In cases where an API produces a this value by transforming its arguments, a new ThisType<T> marker interface can be used to contextually indicate the transformed type. Specifically, when the contextual type for an object literal is ThisType<T> or an intersection including ThisType<T>, the type of this within methods of the object literal is T.

sandersn:
`vue.d.ts` is the one that needs to use `ThisType` I think; I'm not sure what vetur would need to do. @DanielRosenwasser is working on an update right now.

I have the bug almost tracked down -- vue-hackernews works great as long as you don't change one of the files to typescript and then try to open other files :)

sandersn:
Hm, so this is not an easy fix. The problem is that javascriptmode.ts doesn't know the languageId of files that are not opened yet, but the language service still wants to know whether they contain JavaScript or typescript. There are two basic approaches to fix this:

1. Start with an initial *guess*. Right now this causes an assert if the guess is wrong because the file type switches in the language service. 
2. Add files to the language model cache as the language service requests them. This bloats the cache, but arguably it makes vetur itself more like a real language service.

I think (2) is the right solution, but it's more work. I'll do what I can but my next 48 hours are pretty busy with non-work things, so it may be Monday before I have something working.

sandersn:
Turns out the fix is pretty small after all: calling `jsDocuments.get` on a new TextDocument implicitly registers it with the cache.

Unfortunately, because I was working from my laptop, I discovered that path handling is completely broken on Windows. `url.parse(filename).path` is not a complete solution for Windows filenames, which are rooted at `c:/` or whatever. There's an extra preceding `/` and (at least) the ':' is encoded as '%3A'. This has to be a solved problem within VS Code, so I'm going to find out what the existing solution is.

octref:
Are you looking for this?

https://github.com/Microsoft/vscode/blob/master/src/vs/base/common/uri.ts#L126-L145

I'm not against including that as part of the source. I might need it too.

However, overall I think it's better to have a fine-tuned default ts/js config internally, as I imagine most Vue users wouldn't bother to specify js/ts config. (Though reading tsconfig if it's present doesn't hurt).

sandersn:
Yes, I am also looking at it right now and wondering what the public version of it is. I asked @mjbvz as well so I should know pretty soon.

Unfortunately, the host also calls `ts.sys.readFile` several times so that the language service is able to analyse files that are not open in the editor. It's not just for tsconfig support. 😿

sandersn:
Looks this package is the right answer: https://github.com/Microsoft/vscode-uri

sandersn:
Ok, windows paths should be working now. Both URL encoding and slash normalisation had to be fixed. I'll check once more on my Linux box to make sure I didn't break anything there, but I think it's ready to merge now.

sandersn:
All right, everything looks fine on Linux as well, so this PR is good to go.

octref:
Thanks a lot! Will include this in 0.6 release.

octref:
@sandersn Do you know @DanielRosenwasser's progress of the `vue.d.ts` update that would use `contextualThis` to transform the wrapped `new Vue()` object?

Or is is I should just start writing a dts lib that would transform a `new Vue()` object into an extended form which would have all `props`, `methods` and `computed` bound to it? Then I can put it in here to bind the new object to `this`: https://github.com/octref/vetur/blob/master/server/src/modes/typescriptMode.ts#L68-L74

DanielRosenwasser:
Right now you can try it all out on https://github.com/DanielRosenwasser/typescript-vue-todomvc, but I need to chat with @yyx990803 about it, but I think the earliest we could get it out is Vue 2.4.

By the way, right now TypeScript users can't rely on contextually typing the object literal because *at compile time*, there is no call to `Vue.extend` (or `new Vue`). So while you'll get no errors in JS scenarios, TypeScript will spit out errors when you run Webpack.

In short, TypeScript users still need to wrap their components in calls to `Vue.extend` which I think is fine.

octref:
> Right now you can try it all out on https://github.com/DanielRosenwasser/typescript-vue-todomvc

Working great for me, thanks!

> TypeScript users still need to wrap their components in calls to Vue.extend

That's fine. I'm wondering if it's possible to modify the source at compile time using a TS loader for webpack.
I didn't delve into vue's internal yet, but I guess `Vue.extend` is the same thing as `class App extends Vue`? People are already writing
```ts
export default class App extends Vue {
  ...
}
```
with https://github.com/vuejs/vue-class-component so I guess that's fine.

Also wondering what's your thoughts on typed props. I see some work towards that in `vue.d.ts`.

Meanwhile I'll start building support for dynamic IntelliSense for template completion in #145. That seems doable without contextual typing anyway.

DanielRosenwasser:
> Also wondering what's your thoughts on typed props

Props is pretty crazy to type because it might come in as an array of strings, or it might come in as an object which you can infer from. Ideally I could have a `PropNames extends string = never` and `Props`, and have `props: SomeMappedType<Props> | PropNames[]` (or something like that), but it started getting complicated.

I'll try it out after next week. Right now we're getting some stuff together for our Build conference.

Issue #136, opened 2017-04-21
Master Issue for Language Server Crash


octref:
First, sorry for the LS crashes.

Here is 0.5.6 in case people need to do work: [vetur-0.5.6.vsix.zip](https://github.com/octref/vetur/files/947389/vetur-0.5.6.vsix.zip).  
Instruction: https://code.visualstudio.com/docs/editor/extension-gallery#_install-from-a-vsix

I'll be traveling during weekend, so I don't think all crashes can be resolved until Monday.

With that being said, here is a guideline for reporting server crashes to help me better collect info and solve the issue:

Please provide the following info for crashes:

1. vetur version
2. error log
3. **a minimal repro case**, which
    - should be either a single file or a small project generated by https://github.com/vuejs/vue-cli
    - tweak it until it crashes vetur's LS
    - post the code on GH issue (for single file) or upload your project to GH and post link (for project)

3 is really important, because without a repro I can't do anything.

Also @sandersn, can you take a look at #127 #128 #131 #135? They seem to come from the module resolution code in your PR.

sandersn:
#128 comes from the hidden, inserted `import Vue from 'vue'` conflicting with an `import Vue` written by the user. (Based on my knowledge of the code &mdash; I haven't reproed it yet.)

sandersn:
#127 and #135 seem to be caused by incorrect setup of vetur's dependency on vue-template-compiler. It is likely that I didn't set up this dependency correctly, which is why @MWallin was able to work around the problem by installing vue-template-compiler as part of their own package.

octref:
Yea I also see #128 coming from there. As for #135, @MWallin seems to be using `vue-template-compiler` as a devDeps and that somehow have an effect on vetur's LS.

sandersn:
Looks like `package.json` at root needs to add a dependency on `vue-template-compiler` according to https://code.visualstudio.com/docs/extensionAPI/extension-manifest

sandersn:
It's probably as simple as adding 

```json
  "dependencies": {
    "vue-template-compiler": "^2.2.1"
  },
```

in the middle of the root `package.json`. I'm not sure how to test it though.

octref:
Also, think we should limit the files to do module resolution for limited to only `.vue` file.

> Looks like package.json at root needs to add a dependency on vue-template-compiler according to https://code.visualstudio.com/docs/extensionAPI/extension-manifest

[...]

Issue #135, opened 2017-04-21
0.6.1 - Vue packages version mismatch


frankdugan3:
After upgrading to 0.6.1, I am having a different issue than the other ones posted thus far. The language server crashes with this error stating the the `vue` and `vue-template-compiler` are mismatched versions. I also tried removing and reinstalling the extension to update the dependencies to match, but the mismatch still happens, so it must be an issue with a `package.json` file somewhere.

```
/home/frankdugan3/.vscode/extensions/octref.vetur-0.6.1/client/server/node_modules/vue-template-compiler/index.js:8
  throw new Error(
  ^

Error: 

Vue packages version mismatch:

- vue@2.1.10
- vue-template-compiler@2.2.6

This may cause things to work incorrectly. Make sure to use the same version for both.
If you are using vue-loader@>=10.0, simply update vue-template-compiler.
If you are using vue-loader@<10.0 or vueify, re-installing vue-loader/vueify should bump vue-template-compiler to the latest.

    at Object.<anonymous> (/home/frankdugan3/.vscode/extensions/octref.vetur-0.6.1/client/server/node_modules/vue-template-compiler/index.js:8:9)
    at Module._compile (module.js:556:32)
    at Object.Module._extensions..js (module.js:565:10)
    at Module.load (module.js:473:32)
    at tryModuleLoad (module.js:432:12)
    at Function.Module._load (module.js:424:3)
    at Module.require (module.js:483:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/home/frankdugan3/.vscode/extensions/octref.vetur-0.6.1/client/server/modes/typescriptMode.js:5:31)
    at Module._compile (module.js:556:32)
[Error - 9:12:49 AM] Connection to server got closed. Server will not be restarted.
```

octref:
Do you have `vue-template-compiler` as a dep in your repo?

sandersn:
#94 adds a dependency on vue-template-compiler for vetur. It's shrinkwrapped to 2.2.1. I guess that's what is causing the mismatch, but I'm not sure how to resolve it. (I'm also not sure why it says 2.2.6)

I remember seeing that dependencies needed to be treated specially by VS Code extensions, so maybe that is the first place to look.

frankdugan3:
I do have `vue-template-compiler` as a dep in my app, and it's at version `2.2.6` which matches my version of `vue`. Since `vetur` has it's own version and `node_modules` directory, shouldn't it be isolated from mine?

[...]

Issue #127, opened 2017-04-20
Server crash on 0.6.1 `var symbol = container.locals.get(node.text);`


pdanpdan:
```
[Error - 8:21:07 PM] Request textDocument/completion failed.
  Message: Request textDocument/completion failed with message: Cannot read property 'get' of undefined
  Code: -32603 
C:\Users\H195849\.vscode-insiders\extensions\octref.vetur-0.6.1\client\server\node_modules\typescript\lib\typescript.js:23318
                var symbol = container.locals.get(node.text);
                                             ^

TypeError: Cannot read property 'get' of undefined
    at isNameOfExportsOrModuleExportsAliasDeclaration (C:\Users\H195849\.vscode-insiders\extensions\octref.vetur-0.6.1\client\server\node_modules\typescript\lib\typescript.js:23318:46)
    at bindStaticPropertyAssignment (C:\Users\H195849\.vscode-insiders\extensions\octref.vetur-0.6.1\client\server\node_modules\typescript\lib\typescript.js:23396:17)
    at bindWorker (C:\Users\H195849\.vscode-insiders\extensions\octref.vetur-0.6.1\client\server\node_modules\typescript\lib\typescript.js:23079:29)
    at bind (C:\Users\H195849\.vscode-insiders\extensions\octref.vetur-0.6.1\client\server\node_modules\typescript\lib\typescript.js:22974:13)
    at visitNode (C:\Users\H195849\.vscode-insiders\extensions\octref.vetur-0.6.1\client\server\node_modules\typescript\lib\typescript.js:14662:20)
    at Object.forEachChild (C:\Users\H195849\.vscode-insiders\extensions\octref.vetur-0.6.1\client\server\node_modules\typescript\lib\typescript.js:14725:21)
    at bindEachChild (C:\Users\H195849\.vscode-insiders\extensions\octref.vetur-0.6.1\client\server\node_modules\typescript\lib\typescript.js:21771:16)
    at bindChildrenWorker (C:\Users\H195849\.vscode-insiders\extensions\octref.vetur-0.6.1\client\server\node_modules\typescript\lib\typescript.js:21852:21)
    at bindChildren (C:\Users\H195849\.vscode-insiders\extensions\octref.vetur-0.6.1\client\server\node_modules\typescript\lib\typescript.js:21746:17)
    at bind (C:\Users\H195849\.vscode-insiders\extensions\octref.vetur-0.6.1\client\server\node_modules\typescript\lib\typescript.js:22985:21)
```

[...]

PR #142, opened 2017-04-23
Add vue-template-compiler to extension manifest

sandersn:
After consulting with @mjbvz, I think this is the fix for the `vue-template-compiler` bugs linked from #136. Unfortunately, I'm not sure how to make a clean install to test the change.


octref:
Worth giving it a try. I'll publish a version containing this fix to see if it addresses #135 and #127.

Issue #131, opened 2017-04-21
server still crashed on 0.6.1 CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory


exonuclease:
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
 1: node::Abort() [/usr/share/code/libnode.so]
 2: 0x7f23b86632a7 [/usr/share/code/libnode.so]
 3: v8::Utils::ApiCheck(bool, char const*, char const*) [/usr/share/code/libnode.so]
 4: v8::internal::V8::FatalProcessOutOfMemory(char const*, bool) [/usr/share/code/libnode.so]
 5: v8::internal::Factory::NewFillerObject(int, bool, v8::internal::AllocationSpace) [/usr/share/code/libnode.so]
 6: v8::internal::Runtime_AllocateInTargetSpace(int, v8::internal::Object**, v8::internal::Isolate*) [/usr/share/code/libnode.so]
 7: 0x3768afe08206
[Info  - 上午8:27:56] Connection to server got closed. Server will restart.
Fedora25
vscode 1.11.2

soundtina:
still creshed!

> [0421/092928:ERROR:tcp_listen_socket.cc(76)] Could not bind socket to 127.0.0.1:6005
> [0421/092928:ERROR:node_debugger.cc(86)] Cannot start debugger server

byends:
+1024

byends:
> workbench.main.js:29 Connection is closed.: Error: Connection is closed.
    at ConnectionError.Error (native)
    at new ConnectionError (C:\Users\BYENDS\.vscode\extensions\octref.vetur-0.6.1\client\node_modules\vscode-jsonrpc\lib\main.js:116:28)
    at throwIfClosedOrDisposed (C:\Users\BYENDS\.vscode\extensions\octref.vetur-0.6.1\client\node_modules\vscode-jsonrpc\lib\main.js:481:19)
    at Object.sendNotification (C:\Users\BYENDS\.vscode\extensions\octref.vetur-0.6.1\client\node_modules\vscode-jsonrpc\lib\main.js:522:13)
    at C:\Users\BYENDS\.vscode\extensions\octref.vetur-0.6.1\client\node_modules\vscode-jsonrpc\lib\main.js:627:32
    at e.invoke (c:\Program Files (x86)\Microsoft VS Code\resources\app\out\vs\workbench\node\extensionHostProcess.js:4:16423)
[...]

PR #161, opened 2017-04-26
Exclude node_modules in module resolution

octref:
/cc @sandersn 

For #131 and alike, this seems to happen when a project is large.

In one of my project I was able to repro it, and while debugging I found:

https://github.com/octref/vetur/blob/master/server/src/modes/javascriptMode.ts#L50-L67
```ts
  const { createLanguageServiceSourceFile, updateLanguageServiceSourceFile } = createUpdater();
  (ts as any).createLanguageServiceSourceFile = createLanguageServiceSourceFile;
  (ts as any).updateLanguageServiceSourceFile = updateLanguageServiceSourceFile;
  const configFilename = ts.findConfigFile(workspacePath, ts.sys.fileExists, 'tsconfig.json') ||
    ts.findConfigFile(workspacePath, ts.sys.fileExists, 'jsconfig.json');
  const configJson = configFilename && ts.readConfigFile(configFilename, ts.sys.readFile).config || {};
  const parsedConfig = ts.parseJsonConfigFileContent(configJson,
    ts.sys,
    workspacePath,
    compilerOptions,
    configFilename,
    undefined,
    [{ extension: 'vue', isMixedContent: true }]);

  // This files is an array of over 6k items with all js/ts/vue files in current workspace, including those in node_modules
  const files = parsedConfig.fileNames; 

  compilerOptions = parsedConfig.options;
  compilerOptions.allowNonTsExtensions = true;
```

Then later in here:

https://github.com/octref/vetur/blob/master/server/src/modes/javascriptMode.ts#L89-L109
```ts

    resolveModuleNames(moduleNames: string[], containingFile: string): ts.ResolvedModule[] {
      // This is run 6k times, one time for each file
      return moduleNames.map(name => {
        if (path.isAbsolute(name) || !isVue(name)) {
          return ts.resolveModuleName(name, containingFile, compilerOptions, ts.sys).resolvedModule;
        }
        else {
          const uri = Uri.file(path.join(path.dirname(containingFile), name));
          const resolvedFileName = uri.fsPath;
          if (ts.sys.fileExists(resolvedFileName)) {
            const doc = docs.get(resolvedFileName) ||
              jsDocuments.get(TextDocument.create(uri.toString(), 'vue', 0, ts.sys.readFile(resolvedFileName)));
            return {
              resolvedFileName,
              extension: doc.languageId === 'typescript' ? ts.Extension.Ts : ts.Extension.Js,
            };
          }
        }
      });
    },
```

This function is called for each files. So for the whole worksapce it's called 6k times, one time for each js/ts/vue file, and then for each file TS was resolving all modules in them...

I don't know if this PR is the right approach, but it seems to solve the problem.

@sandersn 
Meanwhile in https://github.com/octref/vetur/issues/136#issuecomment-296257118 you said
> I think there may be some tsconfig settings to limit how deeply the compiler follows module resolution inside `node_modules`.

Do you mean `maxNodeModuleJsDepth`? That might help solving this issue, but how is the depth calculated?

### DIFF
--- a/server/src/modes/javascriptMode.ts
+++ b/server/src/modes/javascriptMode.ts
@@ -54,7 +54,9 @@ export function getJavascriptMode(documentRegions: LanguageModelCache<HTMLDocume
   (ts as any).updateLanguageServiceSourceFile = updateLanguageServiceSourceFile;
   const configFilename = ts.findConfigFile(workspacePath, ts.sys.fileExists, 'tsconfig.json') ||
     ts.findConfigFile(workspacePath, ts.sys.fileExists, 'jsconfig.json');
-  const configJson = configFilename && ts.readConfigFile(configFilename, ts.sys.readFile).config || {};
+  const configJson = configFilename && ts.readConfigFile(configFilename, ts.sys.readFile).config || {
+    exclude: ['node_modules']
+  };
   const parsedConfig = ts.parseJsonConfigFileContent(configJson,
     ts.sys,
     workspacePath,


sandersn:
`maxNodeModuleJsDepth` means the number of times to follow `import` within imported `.js` files that are under `node_modules`. So if your code imports 'libraryA', libraryA is at depth 0. When 'libraryA' imports 'libraryB', libraryB is at depth 1, and so on if libraryB imports more modules. Module resolution stops after it hits the max and adds further imports to a list of "elided imports", which basically means that the module is treated as type `any`.

I confirmed with @mhegazy that if not provided it defaults to 0, so it should not make much difference.

This is *different* from excluding node_modules. That stops the compiler from its *normal*, non-import search for source files, which sounds like exactly what we want here. `maxNodeModuleJsDepth` only applies when somebody has explicitly said `import X from 'library'` and the compiler finds `library` inside `node_modules`.

octref:
@sandersn Thanks for the info!

octref:
@sandersn 

Is there anyway to exclude `node_modules` even in subdirectories of workspaces?

Tried:

```
exclude: ['node_modules', '**/node_modules']
```

No effect, still breaks on https://github.com/mrgodhani/rss-reader per https://github.com/octref/vetur/issues/131#issuecomment-297591254. `files` still include everything under `app/node_modules/`.

---

OK, just figured out this works:

```
exclude: ['node_modules', '**/node_modules/*']
```

This seems to match everything under `app/node_modules`. Still, a bit surprising that `**/node_modules` wouldn't match everything under `app/node_modules` -- would you say that's a bug for TS?

sandersn:
I don't actually know the glob code. @rbuckton, can you comment whether the above `**/node_modules/*` is according to spec?

Here are the PRs in case you want to look through the code yourself:
Microsoft/typescript#8841 
Microsoft/TypeScript#5980

PR #199, opened 2017-05-12
Improve documentation about jsconfig.json

qiansen1386:
Hi @octref, as far as I know, we need tsconfig/jsconfig for formatting, is it?


octref:
Thanks for the PR. But I've enabled formatter even for files not in tsconfig/jsconfig: https://github.com/octref/vetur/commit/b695472992ad9dfbd5356cd6235afdced80cfa93

I'll add some clarification though.

PR #281, opened 2017-06-25
fix document error

ThinkRealize:
client code need  compile.


octref:
The idea is `server` has its own `compile`, as we distribute server by publishing it to npm.
The client's `compile` script is hooked to debug / task, see https://github.com/octref/vetur/blob/master/.vscode/tasks.json and https://github.com/octref/vetur/blob/master/.vscode/launch.json

So overall I'm not in for this PR. Next time maybe open an issue for discussion first!

Looking forward to seeing other PRs from you 👍

PR #299, opened 2017-06-28
doc: fix some invalid or redirected url

starkwang:
Some of the urls in README is invalid or redirected. This PR is to fix them.


octref:
Thanks!

PR #329, opened 2017-07-13
Improve support for closing tags in javascript

armano2:


octref:
Umm...Although backstick would affect html / css too I think it should be fine. Thanks!

Issue #377, opened 2017-08-11
Emment not working anymore with VS Code update


DevoidCoding:
## Info

- Platform: Win
- Vetur version: 0.9.3
- VSCode version: 1.15.0

## Problem

With the last update of VS Code (updated yesterday), the emmet aren't working anymore on vue-html.
Here's my settings :
```json
"emmet.useNewEmmet": false,
"emmet.syntaxProfiles": {
    "vue-html": "html",
    "vue": "html"
}
```

Now, the "emmet.useNewEmmet" is an unknow configuration setting

## Reproducible Case

Just create a new file or use older .vue file

DevoidCoding:
Replacing it by
```json
"emmet.includeLanguages": {
            "vue-html": "html",
            "vue": "html"
    }
```
will fix the problem

octref:
`"vue": "html"` only fixes it for html. It doesn't limit it to the `<template>` region or has CSS support.
I'm working on integrating emmet directly and will make a release soon. You can subscribe to #232.

PR #381, opened 2017-08-12
fix emmet for the VS Code 1.15.0

DevoidCoding:
Change emmet doc to use emmet 2.0

https://code.visualstudio.com/blogs/2017/08/07/emmet-2.0


octref:
Per https://github.com/vuejs/vetur/issues/377#issuecomment-321974230, this change would only work partially. I'd rather integrate the new emmet and then update the doc. Thanks!

Issue #680, opened 2018-02-04
Make scaffold snippet follow `tabSize`


j127:
- [X] I have searched through existing issues
- [X] I have read through [docs](https://vuejs.github.io/vetur)
- [X] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

## Info

- Platform: Ubuntu 16.04
- Vetur version: 0.11.7
- VS Code version: 1.19.3

## Problem

Typing `template` and pressing `TAB` doesn't respect the file indentation settings. Even if I have it set at 4 spaces (both in the file and with `.editorconfig`, it only indents by 2.

![spaces-vetur](https://user-images.githubusercontent.com/96479/35773051-227cded4-08fc-11e8-874d-4f01836c70f5.png)

## Reproducible Case

1. Set a blank `.vue` file to 4 space indentation.
2. Type `template` on the first line and hit `TAB`.

Expected behavior: I want to end up between two `<template>` tags indented by four spaces.
Actual behavior: I end up between two `<template>` tags indented by two spaces.

This is my first day using VS Code. I don't know how VS Code does it, but I think that in Sublime Text 3, if you indent with tabs in the template, ST3 will convert the tabs to whatever indentation is specified for that file.

octref:
First, you need this setting: `"vetur.format.defaultFormatter.html": "js-beautify-html"`

Second, we don't read from `.editorconfig`. Try setting `editor.tabSize` to 4.

j127:
Thanks for the reply. It's already set to 4 spaces (see the screenshot in the bottom right corner). What should `vetur.format.defaultFormatter.html` be set to in order for it to read the correct indentation settings?

octref:
https://vuejs.github.io/vetur/formatting.html#js-beautify-html-deprecated

j127:
Thanks. It says "tabSize and insertSpaces are read from editor.tabSize and editor.insertSpaces." but vetur isn't reading the correct indentation settings. I have it set to 4 but it's only indenting by 2.

octref:
OK...Do you have a global `editor.tabSize` set as 4?

If so, what's happening is you have

```
  "editor.detectIndentation": true
```

which is the default.

When you run this command
`Cmd+Shift+P -> Indent Using Spaces`
It's probably saying 2 spaces.

You can turn it off by setting the `detectIndentation` as false.

j127:
[...]

PR #685, opened 2018-02-08
Make snippets use the editor's indentation settings

misenhower:
This should fix issue #680. The tab character is automatically converted to the current editor's indentation setting (so it's automatically expanded to 4 spaces or whatever the current config is).


octref:
This indeed fixes #680. Thanks!

PR #806, opened 2018-05-28
Remove the video link because it's broken

flaviocopes:
Youtube says the video is no longer available 😔


octref:
This is an accident and @shentao will take care of it.

@shentao Once you resolve the issue, do you mind replying with an updated URL that would work? Thanks.

shentao:
The links should be working again.  :)

PR #846, opened 2018-07-26
fix: prettier-eslint no inferred parser error (fix #835)

### LINKED ISSUE #835: File not format

## Info

- Platform: Windows
- Vetur version: 0.12.5
- VS Code version: 1.24.1

## Problem

The `.vue` file is not formated.
I have this code :
```vue
<template>
    <h1>Hello World</h1>
</template>

<script>
import seo from  "~/plugins/seo"

export default {
  head() {
    return {
      title: "Accueil",
      meta: [
        ...seo.title("Accueil")
      ]
    }
  }
}
</script>
```
And my eslint :
```js
module.exports = {
  root: true,
  env: {
    browser: true,
    node: true
  },
  extends: [
    // https://github.com/vuejs/eslint-plugin-vue#priority-a-essential-error-prevention
    // consider switching to `plugin:vue/strongly-recommended` or `plugin:vue/recommended` for stricter rules.
    'plugin:vue/recommended',
    'standard'
  ],
  // required to lint *.vue files
  plugins: ['vue'],
  // add your custom rules here
  rules: {}
}
```

I wait this code formated : 
```vue
<template>
  <h1>Hello World</h1>
</template>

<script>
import seo from '~/plugins/seo'

export default {
  head () {
    return {
      title: 'Accueil',
      meta: [
        ...seo.title('Accueil')
      ]
    }
  }
}
</script>
```

The error in console :
```
Vetur initialized
[Error - 14:54:25] prettier-eslint [ERROR]:
Prettier-Eslint format failed
[Error - 14:54:25] No parser could be inferred for file: d:\DEV\NodeJS\clients\fast-luxury-service\v2\pages
[Error - 14:54:25] (node:16584) DeprecationWarning: The 'parserOptions.ecmaFeatures.experimentalObjectRestSpread' option is deprecated. Use 'parserOptions.ecmaVersion' instead. (found in "standard")
```

## Reproducible Case

1. Create project from Veturpack
2. install packages :
```sh
yarn add --dev eslint eslint-config-standard eslint-friendly-formatter eslint-loader eslint-plugin-import eslint-plugin-node eslint-plugin-promise eslint-plugin-standard eslint-plugin-vue babel-eslint
```
3. rewrite `.eslintrc` to
```json
{
  "root": true,
  "env": {
    "es6": true,
    "browser": true,
    "node": true
  },
  "parserOptions": {
    "parser": "babel-eslint"
  },
  "extends": [
    "plugin:vue/recommended",
    "standard"
  ],
  "plugins": ["vue"],
  "rules": {
    "vue/html-self-closing": "off",
    "quotes": ["error", "double"]
  }
}
```

And now when we format `.vue` like `Counter.vue` the file is not formatted /wrong format.

### OR

1. Create project from `nuxt-starter-template` (https://github.com/nuxt-community/starter-template)
2. Try format an `.vue` file (like `/pages/index.vue`) and here the file is not format, this generate the same error of my problem.

## Suggestion 

I have try fix this bug and probably we can use prettier-eslint v8.8.2 and use pass all .vue file for fix it.

**UPDATE** : I have find an solution (https://alligator.io/vuejs/eslint-vue-vetur/) but i keep this issue open because this is not right to make that for can use it

Thanks

zyy7259:
# Code Review

(You can use the link to view the code)
(Code snippet are also pasted below)

- [Currently](https://github.com/vuejs/vetur/blob/master/server/src/utils/prettier/index.ts#L43), `vetur` pass the folder of code as the `filePath`.
- [And](https://github.com/prettier/prettier/blob/cdcb7b11c9afc850ab24420407144772983eefee/src/main/options.js#L44) so `prettier` is going to use that `filePath` to infer a proper parser.
- Since the `filePath` is actually a folder. [The](https://github.com/prettier/prettier/blob/65d595d60aa6b7d5e8798c2a5fca962f2305c196/src/main/options.js#L115) `extension` will be `''`.
- [And `prettier`](https://github.com/prettier/prettier/blob/65d595d60aa6b7d5e8798c2a5fca962f2305c196/src/main/options.js#L123) won't able to infer a proper parser from an empty extension.
- [So](https://github.com/prettier/prettier/blob/65d595d60aa6b7d5e8798c2a5fca962f2305c196/src/main/options.js#L45) `prettier` will throw an error `No parser could be inferred for file`.

```js
    const prettierifiedCode = prettierEslint({
      filePath: require('path').dirname(filePath),
      text: code,
      fallbackPrettierOptions: prettierOptions
    });
```

```js
  if (!rawOptions.parser) {
    if (!rawOptions.filepath) {
      const logger = opts.logger || console;
      logger.warn(
        "No parser and no filepath given, using 'babylon' the parser now " +
          "but this will throw an error in the future. " +
          "Please specify a parser or a filepath so one can be inferred."
      );
      rawOptions.parser = "babylon";
    } else {
      rawOptions.parser = inferParser(rawOptions.filepath, rawOptions.plugins);
      if (!rawOptions.parser) {
        throw new UndefinedParserError(
          `No parser could be inferred for file: ${rawOptions.filepath}`
        );
      }
    }
  }
```

```js
function inferParser(filepath, plugins) {
  const filepathParts = normalizePath(filepath).split("/");
  const filename = filepathParts[filepathParts.length - 1].toLowerCase();

  const language = getSupportInfo(null, {
    plugins
  }).languages.find(
    language =>
      language.since !== null &&
      ((language.extensions &&
        language.extensions.some(extension => filename.endsWith(extension))) ||
        (language.filenames &&
          language.filenames.find(name => name.toLowerCase() === filename)))
  );

  return language && language.parsers[0];
}
```

# Quick fix

- Don't pass `filePath` [and](https://github.com/prettier/prettier/blob/65d595d60aa6b7d5e8798c2a5fca962f2305c196/src/main/options.js#L41) `prettier` will use the default parser `babylon `.

# PS

According to `prettier`'s [code](https://github.com/prettier/prettier/blob/65d595d60aa6b7d5e8798c2a5fca962f2305c196/src/main/options.js#L33)

```
        "No parser and no filepath given, using 'babylon' the parser now " +
          "but this will throw an error in the future. " +
          "Please specify a parser or a filepath so one can be inferred."
```

So this is just a temporary fix. Since `prettier-eslint` can handle `.vue` files, I think a more resonable fix would be

- pass the vue file's content as text and the corresponding filePath

But this will require a lot of reconstruction and also I don't know if the performance is OK. So I just go with the quick fix.

### DIFF
--- a/server/src/utils/prettier/index.ts
+++ b/server/src/utils/prettier/index.ts
@@ -40,7 +40,9 @@ export function prettierEslintify(
     const prettierOptions = getPrettierOptions(prettierVSCodeConfig, parser, filePath);
 
     const prettierifiedCode = prettierEslint({
-      filePath: require('path').dirname(filePath),
+      // filePath will lead to Error: No parser could be inferred for file
+      // maybe pass the vue file's content as text and the corresponding filePath?
+      // filePath: require('path').dirname(filePath),
       text: code,
       fallbackPrettierOptions: prettierOptions
     });


zyy7259:
The CI build failed with a `Stylus` check :(

zyy7259:
More about the default parser problem in this issue: https://github.com/prettier/prettier/issues/4585

zyy7259:
I run `yarn test` locally. There is no `Stylus Completion - basic property` error.

image

zyy7259:
After a rebase, there seems no error anymore

Coder-256:
@octref This is still an issue. See the discussion on prettier/prettier-vscode#498. Is there any reason this should not be merged?

octref:
Sorry that it took me so long to get back to you.

I think the solution should be to ensure filepath passed to prettier/prettier-eslint always end up with the js/ts suffix, instead of this fix.

octref:
If you want to continue the PR, please do:
- Give filepath the correct extensions before passing it to prettier/prettier-eslint
- For a given vue file like `foo/bar.vue`, see if passing `foo/bar.js` to prettier would work fine (since only `bar.vue` exists on disk, not `bar.js`.

PR #855, opened 2018-08-01
Update README.md

chemzqm:
LanguageClient-neovim throws with current configuration. 
LanguageClient-neovim doesn't works well with completion of css, html and json.

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -50,3 +50,13 @@ There are two ways to integrate `vue-language-server` into editors:
     }
   }
   ```
+
+3. As extension of [coc.nvim](https://github.com/neoclide/coc.nvim)
+
+  Install [coc.nvim](https://github.com/neoclide/coc.nvim) in your vim/neovim.
+
+  Then, run vim command
+
+  ```
+  :CocInstall coc-vetur
+  ```


octref:
I'm neutral to using either. @HerringtonDarkholme as you use NeoVim more, what's your thoughts on this one?

Also what's the difference between your client and LanguageClient-Neovim?

chemzqm:
Here're some benefits for the user I could think of: 

* Easier to install, no requirement for other plugins, prompt user for download if module not found.
* Easier to debug with support of `node-rpc` and output channel https://github.com/neoclide/coc.nvim/wiki/Debug-language-server
* More flexible, user could have multiply language server for single filetype.
* Well supported for all completion features defined in LSP specification
* Support file watching, custom commands, middleware features for extensions.

The disadvantage is the binary of coc is much bigger thanLanguageClient-Neovim.

octref:
However, COC seems to be only for completion whereas LanguageClient-NeoVim has support for many other language features. As VLS is not limited to completion, I think LanguageClient-Neovim would be more appropriate to put there as an example.

chemzqm:
>  COC seems to be only for completion

That's not true, coc have builtin extensions, while user could have their own configuration to support any language server.  COC comes with completion support to correctly support completion of LSP, while many features like completion resolve request is not supported by LanguageClient-neovim

chemzqm:
@HerringtonDarkholme 👌

chemzqm:
Updated, coc now use node module as extension.

octref:
Excellent, I might give it a try. I have to use VS Code for work but I miss Vim a lot.

Issue #873, opened 2018-08-13
Vetur 1.0 Roadmap


octref:
## Synopsis

A lot of work lies ahead to get Vetur to 1.0. This roadmap exists to:

- Give a big picture as to what's left.
- Inform the complexity of the remaining issues.
- Inform the prioritization of features/bug-fixes.
- Guide you to the correct issue to follow up.

This issue is a live document updated frequently.
This issue supersedes #188.

### Signage

- :star2: Current focus.
- :running_man: In progress.
- :rotating_light: Upstream issue or support needed.
- ☑️: Features to implement.

### Discussion in This Issue

Please make the discussion stay high-level.

Good Discussion: 👍
- *I think this feature area is missing, and it's important because...*
- *If you have xxx, it would make it much easier for people to contribute...*

Bad Discussion: 👎
- *When can you support pug formatting?* (Complain in individual issues).
- *How do I configure Vetur to do xxx?* (Open new issues for questions).

## Features

Here is how Vetur looks like from each feature area. In each section, I describe:

- Current state
- What's missing
- Complexity of the features

This section provides a high-level overview and omits smaller bugs.

Vote for areas that you want more support:

[![](https://api.gh-polls.com/poll/01CMRCGRJ1PVCD604PPHD9BGMH/Smart%20Editing%20in%20Template%20Interpolations)](https://api.gh-polls.com/poll/01CMRCGRJ1PVCD604PPHD9BGMH/Smart%20Editing%20in%20Template%20Interpolations/vote)
[![](https://api.gh-polls.com/poll/01CMRCGRJ1PVCD604PPHD9BGMH/Multi-root%20and%20Sub-folder%20support)](https://api.gh-polls.com/poll/01CMRCGRJ1PVCD604PPHD9BGMH/Multi-root%20and%20Sub-folder%20support/vote)
[![](https://api.gh-polls.com/poll/01CMRCGRJ1PVCD604PPHD9BGMH/JSX%20Support)](https://api.gh-polls.com/poll/01CMRCGRJ1PVCD604PPHD9BGMH/JSX%20Support/vote)
[![](https://api.gh-polls.com/poll/01CMRCGRJ1PVCD604PPHD9BGMH/TypeScript%20Plugin%20Support)](https://api.gh-polls.com/poll/01CMRCGRJ1PVCD604PPHD9BGMH/TypeScript%20Plugin%20Support/vote)
[...]

PR #876, opened 2018-08-15
prefer file local prettier

maeldur:
I brought in the [file from prettier-vscode](https://github.com/prettier/prettier-vscode/blob/master/src/requirePkg.ts) to have vetur [behave the same way as that extension](https://github.com/prettier/prettier-vscode#prettier-resolution)   

I've tested this change with no prettier, an older version of prettier, and a newer one and it seems to work ok.. is there more formal testing i should do?


maeldur:
Btw, thanks for the awesome extension and really appreciate the work you did in #873! hopefully this takes us to 1/15 😄

octref:
Sorry for the late response. Mostly look good. Once you make the requested changes I'll pull it in.

maeldur:
Hi @octref, changes made.

re: copyright, I added an entry into ThirdPartyNotices with the LICENSE from prettier-vscode. is that what you meant?  If not, let me know if there's something else I can do.

octref:
@maeldur Thanks for adding it to the ThirdPartyNotice. In the `requirePkg.ts` file's top, please also include something like (use `y` to permalink to the file in a specific commit)

```ts
/**
 * Adapted from https://github.com/prettier/prettier-vscode/blob/1843acb5defac7898862a1df61cb67c7a8355d69/src/requirePkg.ts
 */
```

maeldur:
@octref thanks for the review - added. i went 9 digits into the sha for the blob but if you prefer the long one i'll use the whole thing.

octref:
Thanks again. I just made a release. I'll try to bring in prettyhtml and make a new release again.

PR #912, opened 2018-09-25
Prettyhtml. Fix #561. Fix #491

### LINKED ISSUE #491: prettier for Vue
We already have the infra for splitting Vue component into multiple regions, and currently we feed

- `<template>`: none
- `<style>`: prettier with parser `css` / `scss` / `less`
- `<script>`: prettier with parser `babylon` / `typescript`

Instead of waiting for prettier's html formatter, https://github.com/reshape/reshape by @jescalan seems to be in better shape already. With two more things, I think I can put it in as the default html formatter for Vetur:

- `printWidth`. This is especially important for wrapping html attributes. eg:

  ```html
  <div attr="foo" attr="foo" attr="foo" attr="foo" attr="foo" attr="foo" attr="foo" attr="foo" >bar</div>
  ```
  to

  ```html
  <div
    attr="foo"
    attr="foo"
    attr="foo"
    attr="foo"
    attr="foo"
    attr="foo"
    attr="foo"
    attr="foo"
  >bar</div>
  ```
- Handling of interpolation in attributes and mustache brackets. Example from Vue doc: https://vuejs.org/v2/guide/class-and-style.html#Object-Syntax

  ```html
  <div class="static"
    v-bind:class="{ active: isActive, 'text-danger': hasError }">
  </div>
  ```

  The class object can be very long and needs to be formatted to multiline.

Another thing we could do is to make a CLI based on prettier + reshape + [vue-template-compiler](https://www.npmjs.com/package/vue-template-compiler). Basically just using `vue-template-compiler` instead of Vetur's region splitting.

```
vue component -> vue-template-compiler
  - template -> reshape
  - style -> prettier
  - script -> prettier
    -> assemble into prettierified vue component
```

@jescalan Do you have any thoughts? Would you be interested in working together on this?

### LINKED ISSUE #561: New html formatter
Hey guys!

Because js-beautify-html was driving me nuts :grimacing:, I started working on a html formatter based on [reshape parser](https://github.com/reshape/parser). I just wanted to let you know, so that we don't duplicate out efforts.

I currently have a proof of concept, I'll setup a page shortly in which we can test it by live formatting entered html.

octref:
@starptech I ran it through a few real-world examples. Overall looking pretty good!

If you can handle https://github.com/Prettyhtml/prettyhtml/issues/29 and https://github.com/Prettyhtml/prettyhtml/pull/28 I believe it's good to go.

As for the options, I would suggest `usePrettier` -> `formatEmbeddedContent`. It's not a deal breaker for me but hope it can be more explicit.

Todos for myself:

- Add unit tests
- Add E2E tests using real-world examples

### DIFF
--- a/docs/formatting.md
+++ b/docs/formatting.md
@@ -6,9 +6,14 @@ Vetur has support for formatting embedded `html/css/scss/less/postcss/stylus/js/
 **As a result, only `Format Document` command will be available.**  
 **`Format Selection` command wouldn't work, and there is no**
 
-**html formatting is disabled by default. See [js-beautify-html](#js-beautify-html-deprecated).**
-
 ## Formatters
+
+These formatters are available:
+
+- [`prettier`](https://github.com/prettier/prettier): For css/scss/less/js/ts.
+- [`prettyhtml`](https://github.com/Prettyhtml/prettyhtml): For html.
+- [`stylus-supremacy'](https://github.com/ThisIsManta/stylus-supremacy): For stylus.
+- [`vscode-typescript`](https://github.com/Microsoft/TypeScript): For js/ts. The same js/ts formatter for VS Code.
 
 Choose each language's default formatter in VS Code config, `vetur.format.defaultFormatter`.
 **Setting a language's formatter to `none` disables formatter for that language.**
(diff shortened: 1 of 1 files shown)


StarpTech:
Hi @octref I just released an overwhelmed version of prettyhtml. Please upgrade to the latest version. This includes a new parser which is less strict and much easier to maintain. The issues above should be already gone.

PR #924, opened 2018-10-06
Fix example code

tu4mo:


octref:
How did I end up having that code over there — thanks for the PR!

PR #925, opened 2018-10-06
upgrade prettyhtml

StarpTech:
@octref This PR will upgrade prettyhtml to 0.2.0. It contains fixes for the ignore directive `<!--prettyhtml-ignore-->`

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -21,7 +21,7 @@
   },
   "homepage": "https://github.com/vuejs/vetur/tree/master/server",
   "dependencies": {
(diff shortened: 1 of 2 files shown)


octref:
You need to use `yarn add` instead of NPM. This will update the `yarn.lock` file too.

StarpTech:
Done. Any reason why the node version is stick to `~8` ?

octref:
Just noticed that. I did a upgrade of all deps in `/server` and somehow yarn rewrote it from `>=6` to `~8`. I'll change it back.

PR #927, opened 2018-10-08
upgrade prettyhtml due to bugfix

StarpTech:
https://github.com/Prettyhtml/prettyhtml/releases/tag/v0.2.2

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -21,7 +21,7 @@
   },
   "homepage": "https://github.com/vuejs/vetur/tree/master/server",
   "dependencies": {
-    "@starptech/prettyhtml": "^0.2.1",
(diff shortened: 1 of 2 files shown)


octref:
Thanks for the PR. However I think it would be a better model for me to upgrade dependencies all together before release. That's the same way I upgrade stylus-supremacy, and you don't need to send a PR for each release.

StarpTech:
Sure, I report it because it's a very important release. It will fix an issue which can be occurred every time and the formatter will fail.

octref:
I see, I'll try to make a release this week then, before which I'll update the dependencies.
Thanks for keeping me posted!

Issue #695, opened 2018-02-18
Vetur Prettier ignores ESLint rules with prettier.eslintIntegration set to true


SeanGroff:
- [x ] I have searched through existing issues
- [ x] I have read through [docs](https://vuejs.github.io/vetur)
- [ x] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

## Info

- Platform: macOS
- Vetur version: 0.11.7
- VS Code version: 1.20.1

## Problem

Vetur's version of prettier does not respect my ESLint rules even though I have `prettier.eslintIntegration` set to `true`.

 Cannot get `.vue` files to format according to `eslintrc.js`. .js files format without issue.
This causes a problem when using vue-cli and the Standard ESLint rules.

kleinfreund:
This did work as intended before. Was some default changed? I spent half an hour chasing this down.

octref:
Which rule? Also please include a repro case (or repository) and screenshot.

KennyWho:
rules: _space-before-fucntion-paren_
I use prettier-eslint instead of prettier.when it use in _.js_ file, it's right,but in _.vue_ file, it can't match the rules.

iulo:
same issue. Any update?

stolfgang:
prettier.eslintIntegration stops working if I install and activate eslint-config-standard (https://github.com/standard/eslint-config-standard).
To reproduce clone veturpack repro, run
`npm install --save-dev eslint-config-standard eslint-plugin-standard eslint-plugin-promise eslint-plugin-import eslint-plugin-node`
and add "standard" to .eslintrc

Afterwards it ignores eslint settings, but works well on javascript files.

shangxinbo:
I got the same problem. And the vscode show me Unknown configuration setting for "prettier.eslintIntegration". What's up?Because of that, I should config prettier in config.json and both in eslintrc.js

orblazer:
Hello, i have the same problem but me the HTML not want format the HTML and i can't change pretty rules.

My config : https://gist.github.com/orblazer/df1983339bfee462b8f15bfae9656ac3

Thanks.

syropian:
I'm running into this exact issue as well. I'm using:
```
...
extends: ['standard', 'plugin:vue/essential'],
...
```

[...]

PR #929, opened 2018-10-11
Settings for vscode, prettier, prettier-eslint integration

ozum:
There are lots of questions in issues part of vetur and vscode-prettier how to format `template`, `script` and `style` sections of vue files at the same times. Providing example setting in doc greatly helps in those cases.

Original suggestion belongs to @Bartozzz, see: https://github.com/vuejs/vetur/issues/695#issuecomment-428654748


ozum:
According to @octref (https://github.com/vuejs/vetur/issues/695#issuecomment-429034831) this settings are wrong. Please do not merge until more correct settings are available.

octref:
I won't be merging this because the config is not recommended.

PR #952, opened 2018-11-04
Fixes #862

### LINKED ISSUE #862: More detailed explanation on how to integrate with editor back-ends

## Info

- Platform: Ubuntu 18.04 x64
- Vetur version:  vue-language-server@0.0.33
- VS Code version: 1.25.1 - 1dfc5e557209371715f655691b1235b6b26a06be - x64

## Problem

The [server readme](https://github.com/vuejs/vetur/tree/master/server) has an example of how to set up vls to work with Neovim, which currently does not work for me. I also couldn't get it to work with [Onivim](https://github.com/onivim/oni).

**Running the command as `vls --stdio` starts the server as usual, but starting the LanguageClient with the supplied example does not.**

---

During the LanguageClient's start, the following error message is thrown:

`Request initialize failed with message: Cannot read property 'vetur' of undefined`.

Is this because I need to set up some kind of configuration file at the workspace's root? I've tried to search for instructions on inputting this configuration through some CLI command but this isn't covered by the LSP specification.

[vueServerMain.ts#L35](https://github.com/vuejs/vetur/blob/9952064eb94c92cb0b15c7890a093ccbf7014d2b/server/src/vueServerMain.ts#L35) seems to be the line I'm trying to reach, but is it even possible to do that outside of the VSCode client interface?

And I'd also like to bring the topic of even if the server initializes with the correct configurations, would the usability aspects also be sorted out? Are the commands for doing operations (linting, reformatting, etc) issued directly through Node or is it done through some kind of VSCode API?

Since I'm interested in this feature, I'd like to help in any way I can.

archSeer:
Applying the fix from https://github.com/vuejs/vetur/issues/862#issuecomment-435629045

All credit goes to @AlexKornitzer

### DIFF
--- a/server/src/vueServerMain.ts
+++ b/server/src/vueServerMain.ts
@@ -31,7 +31,7 @@ connection.onInitialize((params: InitializeParams): InitializeResult => {
   console.log('Vetur initialized');
   const vls = new VLS(workspacePath, connection);
 
-  if (initializationOptions) {
+  if (initializationOptions && initializationOptions.config) {
     vls.configure(initializationOptions.config);
   }
 


octref:
Thanks!

PR #956, opened 2018-11-07
Update js-beautify-html default settings to match Vue style guide

tarikhamilton:
js-beautify-html has had many fixes and is a bit more active. They've fixed the force-expand-multiline formatting, which was the biggest issue, in my opinion, that made it very broken. I updated the docs to reflect this. I think it is fair to let people know it is a viable option, considering it's popularity and size.

I also set the default settings to match the [Vue style guide](https://vuejs.org/v2/style-guide/).

I changed `auto` to `force-expand-multiline` because `auto` never seemed to trigger multiline expansion and the style guide states anything more than one attribute should be expanded. Of course, this can be overridden by the config.

### DIFF
--- a/docs/formatting.md
+++ b/docs/formatting.md
@@ -69,9 +69,9 @@ VS Code's js/ts formatter built on TypeScript language service.
 
 Other settings are read from `javascript.format.*` and `typescript.format.*`.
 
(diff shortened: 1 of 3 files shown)


octref:
That's fine for me. Thanks.

PR #960, opened 2018-11-08
Add mouseenter, mouseleave events in html tags

dimensi:
I don't know why this events not added in standard events on html elements. I always prefer use mouseenter/leave than to mouseover/out.


dimensi:
@octref I think you confused me with someone :) This is my first pr into your project.

Issue #922, opened 2018-10-04
Enable logging from TS Language Service


octref:
Currently, when we have issues like https://github.com/vuejs/vetur/issues/913, we have no easy way to know what's happening on TS Language Service.

We should enable logging (either directly from TS, or on our side's integration by adding logs to file access) before upgrading TS versions.

rskvazh:
Waiting for TS >=3 support, because already using some magical generics :)

HIMISOCOOL:
Hey there team, has anyone made a start on this? I have no knowledge of the language server but I'd love to have a crack at this!

Issue #913, opened 2018-09-26
High cpu usage with 0.12.7


Vorror:
- [X] I have searched through existing issues
- [X] I have read through [docs](https://vuejs.github.io/vetur)
- [X] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

## Info

- Platform: OSX
- Vetur version: 0.12.7
- VS Code version: 1.27.2

## Problem
I'm experiencing very high cpu usage when trying to use intellisense inside .vue files when using 0.12.7. Everything works correctly when I use the 0.12.1 vsix build directly from github.

Here's a gif with 0.12.1
As you can see everything works correctly, and intellisense always works as-you-type.
![2018-09-25 21 55 23](https://user-images.githubusercontent.com/3613230/46052895-51bf6200-c10e-11e8-85cb-37860e238100.gif)

Here's a gif with 0.12.7
My cpu is pegged at 100%, intellisense as-you-type doesn't work(have to manually use ctrl+space) _and_ even with that it takes a long time to return results.

![2018-09-25 21 57 12](https://user-images.githubusercontent.com/3613230/46052980-c1355180-c10e-11e8-89cf-faaf128a9bd7.gif)

Vorror:
I know 0.12.1 is pretty old, but it's the only vsix build available. Everything was fine up until today/yesterday, but I can't confirm what build I was before I autoupdated to the latest(0.12.7)

Vorror:
Current veutr settings

```js
  "vetur.validation.template": false,
  "prettier.eslintIntegration": true,
  "vetur.format.defaultFormatter.js": "vscode-typescript",
  "vetur.format.defaultFormatter.html": "js-beautify-html",
  "vetur.format.defaultFormatterOptions": {
    "js-beautify-html": {
      "wrap_line_length": 120,
      "wrap_attributes": "force-aligned"
    }
  }
```

octref:
I made a 0.12.6 package for you. Can you install it and let me know if this causes problem for you? This way I can narrow down the causes.

This is likely caused by upgrading TypeScript to 3.0.3.

https://github.com/vuejs/vetur/releases/tag/v0.12.6

Vorror:
@octref Yep that seems to be the issue

With 0.12.6 everything works correctly
![2018-09-26 01 42 54](https://user-images.githubusercontent.com/3613230/46059570-d2da2180-c12d-11e8-9f8b-bcc292581e44.gif)

[...]

Issue #786, opened 2018-05-14
Make it possible to install different versions of VLS


octref:
Branching off #750

Currently we don't do any Continuous Release, so until we release on marketplace, people who are bugged with specific issue would have trouble getting the latest fix.

However, most issues people have could be solved by updating fixes checked into VLS. Maybe we can hookup the CI to do daily VLS releases like TypeScript does, and make it possible for user to run a command (VS Code / CLI command) to upgrade VLS.

octref:
Not much benefit...People probably just wanted. #682.

PR #993, opened 2018-11-27
Updated to typescript 2.9.2

edgardmessias:
Save ~6MB size of .vsix


octref:
I need to look into the performance issue first before upgrading TS (see #913). So I would like to tackle:
- #922
- #786 

first before I upgrade TS. Thanks.

PR #1013, opened 2018-12-03
Update eslint-plugin-vue to v5

armano2:
https://github.com/vuejs/eslint-plugin-vue/issues/697

i think vetur should also turn off some of rules, they are going to collide with prettier

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -26,7 +26,7 @@
     "buefy-helper-json": "^1.0.2",
     "element-helper-json": "^2.0.2",
     "eslint": "^5.9.0",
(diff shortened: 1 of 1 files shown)


octref:
We use base and essential. So you are saying they are not recommended as default any more?

armano2:
ahh, i was sure that you are using few more, but in that case it's all good

armano2:
btw. no clue why but mac build failed

michalsnik:
Hey @octref is there anything blocking this PR from being merged, except failing build that is rather not caused by this change?

octref:
Nope, sorry. Will publish with new version soon.

Issue #870, opened 2018-08-12
Integrate https://github.com/nuxt-community/nuxt-helper-json


octref:
Somehow I forgot this...

PR #1026, opened 2018-12-13
Tell to install Vue manually for Intellisense (Nuxt)

Tenrys:
Visual Studio Code's recent Vue.js tutorial made me notice that I didn't benefit from Vetur's auto-completion in `.vue` files, and I figured Vue wasn't installed by Nuxt by default, which I'm using. Might as well tell users here?

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -93,3 +93,7 @@ index.vue
 import a from 'components/a.vue'
 import b from 'components/b.vue'
 ```
+
+### Using Nuxt?
+
+Don't forget to install Vue (and probably Vuex / Vue-router) manually, as Nuxt does not come with it by default, to make sure you can use all of Vetur's features, such as auto-complete / Intellisense.


octref:
What way are you creating Nuxt app? I believe it should install Vue dependencies by default.

Tenrys:
I remember doing it with `yarn create nuxt-app` or something, either I removed it accidentally or that doesn't come with it? 

Sorry if I'm mistaken but I was confused for an hour or two about this

Atinux:
I guess it should be resolved by https://github.com/nuxt/nuxt.js/pull/4524 when @octref will add the check for `@nuxt/vue-app` and `@nuxt/vue-app-edge` :)

octref:
This should be made obsolete by the new Nuxt integration in 0.14.4.

octref:
Also see #870.

PR #1033, opened 2018-12-21
refactor: use @types/eslint

ktsn:
It would be better to use `@types` declaration rather than own declaration 🙂


octref:
Looks good - thanks!

PR #1035, opened 2018-12-25
fix mutiple stylus format bug

KennyWho:
Issues: https://github.com/vuejs/vetur/issues/499

```js
function getEmbeddedDocument(document: TextDocument, contents: EmbeddedRegion[], languageId: string): TextDocument {
  const oldContent = document.getText();
  let result = '';
  // main problem point
  // mutiple same languageId will be break
  for (const c of contents) {
    if (c.languageId === languageId) {
      result = oldContent.substring(0, c.start).replace(/./g, ' ');
      result += oldContent.substring(c.start, c.end);
      break;
    }
  }
  return TextDocument.create(document.uri, languageId, document.version, result);
}
```

I use `getValueAndRange` function to reslove correct inputValue.and `format` function in vscode-stylus-supremacy` has multiple arguments. It will be return right formatted text.

origin-before

origin-after

fixed-after

### DIFF
--- a/server/src/modes/style/index.ts
+++ b/server/src/modes/style/index.ts
@@ -154,7 +154,7 @@ function getStyleMode(
   };
 }
 
-function getValueAndRange(document: TextDocument, currRange: Range): { value: string; range: Range } {
+export function getValueAndRange(document: TextDocument, currRange: Range): { value: string; range: Range } {
   let value = document.getText();
   let range = currRange;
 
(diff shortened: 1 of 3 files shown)


KennyWho:
@octref

octref:
Sorry, but https://github.com/vuejs/vetur/pull/1174 is the more comprehensive fix. The fix will be available in next release.

PR #1060, opened 2019-01-15
feat: can read 'vetur' key in local package.json. 

NickeyLin:
I think IntelliSense of some base components or common components in developing projects is also important. I suggest adding local 'vetur' support，so that developers can reuse vue-based components more efficiently.


octref:
It seems you made a few irrelevant changes. If you are still interested in making this change go into Vetur, please open a new PR with clean commits.

PR #1075, opened 2019-01-31
Add support for nuxt-ts and nuxt-ts-edge

P4sca1:
With [nuxt v2.4.0](https://github.com/nuxt/nuxt.js/releases/tag/v2.4.0), `nuxt-ts` and `nuxt-ts-edge` arrived.
This pull request adds vetur support for both dependencies.


octref:
Thanks!

PR #1077, opened 2019-02-02
substitute workspaceRoot with workspaceFolder

fuafa:
https://code.visualstudio.com/docs/editor/variables-reference#_why-isnt-workspaceroot-documented
Seems the `workspaceRoot` has been deprecated, and using it would crash the attached server.


octref:
Thanks and sorry for the late response.

PR #1080, opened 2019-02-07
Simplified grammar compiling process

Patcher56:
Instead of running two editors and having to install the PackageDev Package for Sublime Text, I wrote a small js script to build the grammars.

We can now use `yarn build:grammar` to convert the .yaml files to .tmLanguage.json files as described in the [official vscode docs](https://code.visualstudio.com/api/language-extensions/syntax-highlight-guide).

Also changed the CONTRIBUTING.md to the latest changes.
Maybe we should add the `build:grammar` script either to the compile script or to the pre-push script.


build_grammar.js
```diff
@@ -0,0 +1,13 @@
...
+const path = require('path')
+
+glob(__dirname + '/syntaxes/*.yaml', {nocase: true}, (err, files) => {
+
+  for (const file of files) {
+    const pathData = path.parse(file)
+    fs.writeFileSync(pathData.dir + '/' + pathData.name + '.tmLanguage.json', JSON.stringify(jsYaml.safeLoad(fs.readFileSync(file))))
+  }
+  console.log('built files', files)
+})
```
octref: Run this file through Prettier https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode

(I'll add it as a pre-commit hook #1089)

syntaxes/vue-html.tmLanguage.json
```diff
@@ -0,0 +1 @@
...ntity.other.attribute-name.id.html"},"2":{"name":"punctuation.separator.key-value.html"}},"patterns":[{"name":"string.quoted.double.html","contentName":"meta.toc-list.id.html","begin":"\"","beginCaptures":{"0":{"name":"punctuation.definition.string.begin.html"}},"end":"\"","endCaptures":{"0":{"name":"punctuation.definition.string.end.html"}},"patterns":[{"include":"source.vue#vue-interpolations"},{"include":"#entities"}]},{"name":"string.quoted.single.html","contentName":"meta.toc-list.id.html","begin":"'","beginCaptures":{"0":{"name":"punctuation.definition.string.begin.html"}},"end":"'","endCaptures":{"0":{"name":"punctuation.definition.string.end.html"}},"patterns":[{"include":"source.vue#vue-interpolations"},{"include":"#entities"}]},{"name":"string.unquoted.html","match":"(?<==)(?:[^\\s<>/'\"]|/(?!>))+","captures":{"0":{"name":"meta.toc-list.id.html"}}}]},"unquoted-attribute":{"match":"(?<==)(?:[^\\s<>/'\"]|/(?!>))+","name":"string.unquoted.html"},"tag-stuff":{"patterns":[{"include":"#vue-directives"},{"include":"#tag-id-attribute"},{"include":"#tag-generic-attribute"},{"include":"#string-double-quoted"},{"include":"#string-single-quoted"},{"include":"#unquoted-attribute"}]}}}
```
octref: Use `JSON.stringify(data, null, 2)` — this gives better diff.

Patcher56:
I reordered the grammar files now using the PackageDev Rearrange function.

It rearranges like this:
- include
- comment
- name
- match
- begin
- beginCaptures
- end
- endCaptures
- patterns
- captures

After reordering I also ran prettier.

build_grammar.js
Patcher56: Done with 794dcba

syntaxes/vue-html.tmLanguage.json
Patcher56: Done with a3f9f9e

octref:
For reordering, I was referring to the `build_grammar` script.

I would added logic to reorder the keys in JSON (after loading from YAML), so we can always diff the JSON easily. Manually rewriting the yaml also works — thanks!

Patcher56:
Thank you for merging!
I also thought of adding a logic for reordering.. I definitely will try that in the future.

PR #1081, opened 2019-02-08
Adding liquid markup

digitalfuel:
Can Liquid markup be added to supported grammar Scopes 
Liquid language Support by Neil Ding - neilding.language-liquid


octref:
There are a lot more changes you needed to do. I implemented it via #1260.

PR #1090, opened 2019-02-13
update prettyhtml to 0.8.16

xiaomingplus:
For apply some new features.


octref:
Sorry for my late response. The master has moved forward but merging `yarn.lock` is a pain, so I decided to update all dependencies myself all together. Still, thanks for your contribution!

Issue #1089, opened 2019-02-13
Make prettier a pre-commit hook 


octref:
This should make PR easier.

Issue #1088, opened 2019-02-11
Custom blocks break the highlighting


msaelices:
- [X] I have searched through existing issues
- [X] I have read through [docs](https://vuejs.github.io/vetur)
- [X] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

## Info

- Platform: macOS
- Vetur version: 0.15
- VS Code version: 1.31.0

## Problem

Custom blocks as `<template-native>` are considered as a template but the `-native` postfix break all the following blocks, like `<script>`, `<style>` etc.

<img width="761" alt="screen shot 2019-02-11 at 12 51 28" src="https://user-images.githubusercontent.com/136875/52561763-a08b4100-2dfc-11e9-8603-49cc05f5f42c.png">

## Reproducible Case

Try this code:

```
<template>
  <div class="w-page">
    <div class="w-container">
      <img src="~/assets/logo.png" alt="logo" height="20%" width="20%">
      <HelloWorld :msg="msg"/>
    </div>
  </div>
</template>

<template-native>
  <Page>
    <ActionBar :title="navbarTitle"/>
    <GridLayout rows="auto, auto">
      <Button text="Home" @tap="goToHomePage" row="0"/>
      <Button text="About" @tap="goToAboutPage" row="1"/>
    </GridLayout>
  </Page>
</template-native>

<script>
  import HelloWorld from 'components/HelloWorld';
  export default {
    name: 'home',
    components: {
      HelloWorld
    },
    data() {
      return {
        msg: 'Mode=' + TNS_APP_MODE + ' and Platform=' + TNS_APP_PLATFORM,
      };
    },
  };
</script>
```

Even using the following Custom Block configuration, it still does not work:

```
    "vetur.grammar.customBlocks": {
      "template-native": "html"
    }
```

DestinyHunter:
I have the same problem 

## Info

- Platform: Win
- Vetur version: 0.15.0
- VS Code version: 1.31.1

![co1u 0en1 ld a ho9rx](https://user-images.githubusercontent.com/20897838/52690168-bc771a00-2f97-11e9-912c-291584f7615d.png)
![5x7o83w 9n3dpqa x wur](https://user-images.githubusercontent.com/20897838/52690181-c862dc00-2f97-11e9-87e9-7027bb98e004.png)

octref:
This:

```json
{
  "vetur.grammar.customBlocks": {
    "docs": "md",
    "i18n": "json",
[...]

PR #1093, opened 2019-02-14
Improved grammar and syntax

Patcher56:
- Added prettier as a pre-commit-hook as described in PR #1080 and #1089 
- Changed Contributing.md in .github folder (which is referenced by README.md)
- Fixed custom block generation (as issued in #1088)


syntaxes/vue.yaml
```diff
@@ -102,11 +102,11 @@ patterns:
           - include: text.pug.slm
 
   # html
-  - begin: (<)(template)(?![^/>]*/>\s*$)
+  - begin: (<)(template)(?=>\s*$)
     beginCaptures:
       '1': { name: punctuation.definition.tag.begin.html }
       '2': { name: entity.name.tag.template.html }
-    end: (</)(template)(>)
+    end: (<\/)(template)(>)
```
octref: Why this? YAML shouldn't need escaping forward slash, right?

client/grammar.ts
```diff
@@ -27,12 +25,13 @@ const SCOPES: { [lang: string]: string } = {
   graphql: 'source.graphql'
 };
 
-export function generateGrammarCommandHandler(extensionPath: string) {
-  const customBlocks: { [k: string]: string } =
-    vscode.workspace.getConfiguration().get('vetur.grammar.customBlocks') || {};
+export async function generateGrammarCommandHandler(extensionPath: string) {
+  const vscode = await import('vscode');
```
octref: What's the reason for making this import and making all these functions async?

client/grammar.ts
```diff
@@ -47,7 +46,7 @@ export function generateGrammarCommandHandler(extensionPath: string) {
   };
 }
 
-function getGeneratedGrammar(grammarPath: string, customBlocks: { [k: string]: string }): string {
+export function getGeneratedGrammar(grammarPath: string, customBlocks: { [k: string]: string }): string {
```
octref: Why export this if it's not used in other places?

syntaxes/vue.tmLanguage.json
```diff
@@ -205,7 +205,7 @@
       ]
     },
     {
-      "begin": "(<)(template)(?![^/>]*/>\\s*$)",
+      "begin": "(<)(template)(?=>\\s*$)",
```
octref: This case would be broken:

```vue
<template functional>

</template>
```

I think we need to look into having tests for the TM grammar...Or at least some fixture files we can test the syntax highlighting on. Otherwise it's risky to do any change on the grammar...

octref:
This case still seems broken:

image

```vue
<template lang="pug">
section
  router-link.header.item(to=`/`)
    img(
      src=`/images/navbar-icon.svg`
      :alt=`title`
      )

  .right.menu
    navbar-menu
</template>
```

syntaxes/vue.yaml
Patcher56: I think, we need to escape the forward slash. If I am not escaping it, the output JSON is like this: `"(</)(template)(>)"` which is not correct regex like shown [here](https://regex101.com/r/IaIjyz/1).

The not escaped forward slash is working I think but I believe it is safer to escape them.

client/grammar.ts
Patcher56: As described in the commit description to 39ab7cd I had to dynamic import the vscode to use the `getGeneratedGrammar` outside of this file (I used it for generating `vue-generated.json` with the `build:grammar` script). Without the dynamic import it throws the error that it cannot find the vscode module.

The alternative would be to put the `generateGrammarCommandHandler` into a seperate file. I can do this if you want.

client/grammar.ts
Patcher56: See comment above.

syntaxes/vue.tmLanguage.json
Patcher56: Yeah, that would be cool. I will look for some test cases.
I thought this is already included in the tests... sorry

Patcher56:
Yeah, that's right. I tried many things but got no good solution - so this case would be a bit trickier.
I would prefer to use a space after the `title` keyword or put it into multiple lines if using the `` ` `` literal.

client/grammar.ts
octref: Oops, sorry I missed the commit descriptions...Was just looking at the final code.

Yeah that's fine for me.

syntaxes/vue.tmLanguage.json
octref: No worries. Sorry but the state of art is we have 0 tests against the grammar.

Don't worry too much about automated testing now. Maybe just create a folder of vue files that should be highlighted correctly and manually check your grammar highlight them fine in editor.

I was thinking about using https://github.com/octref/shiki to assert each token against a specific color for testing.

octref:
If supporting backtick would bring in too much complexity I suggest dropping it, document it in the https://vuejs.github.io/vetur/highlighting.html page and move on.

syntaxes/vue.tmLanguage.json
Patcher56: I will do that in #1096 - thanks!

Patcher56:
Somehow the test for osx failed (succeeded on my osx). @octref can you start the test again please?

syntaxes/vue.yaml
octref: Take a look at https://github.com/octref/textmate-grammar, change `grammar.json` to match `\/` instead of `/`, still the same tokenization result.

vscode-textmate could handle unescaped forward slash. I'm not against the change per-se, but I'd like consistency in the grammar rules (either all `/` or all `\/`). I think it's better to revisit this after we have tests covering various cases.

test/grammar.test.ts
```diff
@@ -0,0 +1,79 @@
...
+  for (const key of keys) {
+    if (d[key] !== p[key]) {
+      return true;
+    }
+  }
+  return false;
+}
+
+describe('colorization', () => {
+  const extensionColorizeFixturePath = 'test/grammar/fixtures';
```
octref:
```suggestion
  const extensionColorizeFixturePath = resolve(__dirname, '../../test/grammar/fixtures');
```

octref:
👍 Thanks a lot for the quick response!

PR #1101, opened 2019-02-18
feat: Add support Gridsome

tyankatsu0105:
# What
Add `gridsome-helper-json` for [Gridsome](https://gridsome.org/)

# Question
I want to autocomplete `<page-query>` and `<static-query>` in the same hierarchy as `<template>`, `<script>` and `<style>`
https://gridsome.org/docs/querying-data
Is this PRs response to the my request ?


octref:
I don't think so, this would add the tags/attributes to the `<template>` area, but not the global area.

From this page I see that `<Layout>` and `<Pager>` are tags inside the `<template>` area. This is what you should be adding to the helper json. For `page-query` and `static-query`, adding them to the JSON file wouldn't help.

tyankatsu0105:
@octref 
Thank you for your kind advice !!
Now I understand the helper json's role.
I changed `gridsome-helper-json`.
https://github.com/tyankatsu0105/gridsome-helper-json/commit/d72408ac0c2e6e6f5161d9620438aba967c3f2bd

I want to autocomplete `g-image`, `g-link` and `Pager` in `template` .
So, Please review my work🙏

octref:
Looks great, thanks a lot!

PR #1110, opened 2019-02-20
Multiple folders test

octref:


octref:
@Patcher56 FYI, the test changes are:

- For any folder under `/test`
- `/test/dir/fixture` will be the workspace folder to test against
- Test files live in `/test/dir` and outside `/test/dir/fixture`

Also this switched to using TS project reference which makes building much faster.

Patcher56:
Nice changes! But I think the latest tests failed although the exit code was 0.

I will take a look.

Issue #1139, opened 2019-03-07
No support for tag interpolation syntax highlighting


nmsmith22389:
- [x] I have searched through existing issues
- [x] I have read through [docs](https://vuejs.github.io/vetur)
- [x] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

## Info

- Platform: Mac
- Vetur version: 0.16.2
- VS Code version: 1.31.1

## Problem

When using the new Pug syntax highlighting tag interpolations aren't highlighted.

<img width="310" alt="screen shot 2019-03-06 at 9 25 09 pm" src="https://user-images.githubusercontent.com/8648114/53930270-59d4e380-4056-11e9-8b59-d37c347fe932.png">

_I should note that regular interpolations also aren't highlighted but that doesn't really matter since I don't think you can even use those with Vue._

cheolhow:
There is also a problem in the pug template that the syntax highlight of the property is the same color as the other html tags.
 In any case, syntax highlighting does not work properly in the pug template.

![스크린샷 2019-03-13 오후 4 02 51](https://user-images.githubusercontent.com/25603134/54259824-55ef1700-45aa-11e9-8e99-8f034dd8b17f.png)

cheolhow:
This screenshot was taken after removing vetur

![image](https://user-images.githubusercontent.com/25603134/54260213-a1ee8b80-45ab-11e9-8640-d5f376e3f18b.png)

octref:
This has been fixed for a while (thanks to contribution from @Patcher56)

![image](https://user-images.githubusercontent.com/4033249/59649521-9bc50800-91b5-11e9-83c2-5b1c5048c127.png)

Issue #1154, opened 2019-03-17
PUG + TS syntax highlighting breaks inside v-for


backbone87:
## Info

- Platform: Win10
- Vetur version: 0.16.2
- VS Code version: 1.32

## Problem

Using pug + typescript breaks syntax highlighting in vue files, when using a TS reserved name inside v-for.

## Reproducible Case

```vue
<template lang="pug">
  ul
    //- syntax highlighting breaks within this line, until eof
    li(v-for="type in types") {{ type }}
</template>

<script lang="ts">
export default {
  data() {
    return { types: ['a', 'b'] };
  },
};
</script>
```

using `v-for="(type) in types"` avoids it.

octref:
This is the same problem as using `type` in HTML interpolation.

The proper fix would be to strip TS-specific keyword constructs from https://github.com/Microsoft/TypeScript-TmLanguage for a JS grammar, but that would need to go into https://github.com/Microsoft/vscode/tree/master/extensions/javascript. I don't currently have time for it.

@Patcher56 You might be interested.

backbone87:
i rather prefer no syntax highlighting for interpolations insteadof broken highlighting, and i guess i am not the only one. maybe add a setting to opt out of interpolation highlighting.

Patcher56:
I also ran into this problem. I just used another keyword instead of ‚type‘ as it is a reserved keyword.

I also currently do not know how I would fix that. I can look if I could create a setting for deactivating highlighting for interpolations in pug.

PR #1162, opened 2019-03-21
Completely renewed pug interpolations

Patcher56:
using vscode syntax highlighting injections now
using original pug syntax with specific injections for vue

could delete whole pug implementation -> only use injections anymore
This should fix every pug syntax highlighting problem with vetur, except the one regarding interpolations and directives (such as #1154)

The advantages are:
- Much less code (cleaner code)
- easier to maintain
- When original pug syntax highlighting updates, this automatically gets updated too
- No own language for vue pug syntax highlighting (injected in the vue language)

This can be done with HTML too, should I do that @octref ?

Also added a commit to support new [v-slot](https://vuejs.org/v2/guide/components-slots.html#Named-Slots-Shorthand) syntax in 5eba764.


Patcher56:
This also fixes #1139

syntaxes/vue.tmLanguage.json
```diff
@@ -76,17 +76,9 @@
...
-        "1": {
-          "name": "punctuation.definition.tag.begin.html"
-        },
-        "2": {
-          "name": "entity.name.tag.template.html"
-        }
-      },
-      "end": "(</)(template)(>)",
-      "endCaptures": {
+      "begin": "(?:^\\s+)?(<)((?i:template))\\b(?=[^>]*lang=[\\\"'](?:pug|jade)[\\\"'])",
```
octref: Not quite sure what is this change for — do you mind explaining your reasoning here?

syntaxes/vue.tmLanguage.json
```diff
@@ -102,17 +94,20 @@
...
-          "contentName": "text.pug.vue-pug",
+          "contentName": "source.vue.pug",
           "begin": "(>)",
           "beginCaptures": {
             "1": {
               "name": "punctuation.definition.tag.end.html"
             }
           },
-          "end": "(?=</template>)",
+          "end": "(?=</(?i:template))",
```
octref: Why are we dropping the `>`?

octref:
Thanks a lot for the change! Reducing code duplication is always good.

> HTML too

Let's do another PR for that once this one gets merged in and people aren't reporting much issues.

syntaxes/vue.tmLanguage.json
Patcher56: Sorry, those two are some leftovers due to some tests before. Reverted it in the last commit.

syntaxes/vue.tmLanguage.json
Patcher56: See comment above.

syntaxes/vue-generated.json
```diff
@@ -162,17 +162,9 @@
...
-      "begin": "(<)(template)\\b(?=[^>]*lang=('jade'|\"jade\"|'pug'|\"pug\"))(?![^/>]*/>\\s*$)",
-      "beginCaptures": {
-        "1": {
-          "name": "punctuation.definition.tag.begin.html"
-        },
-        "2": {
-          "name": "entity.name.tag.template.html"
-        }
-      },
+      "begin": "(<)(template)\\b(?=[^>]*lang=[\\\"'](?:pug|jade)[\\\"'])(?![^/>]*/>\\s*$)",
```
octref: I'd rather not have cleverness here — this allows `'pug"` and `"pug'`, etc.

syntaxes/vue-generated.json
```diff
@@ -162,17 +162,9 @@
...
       "end": "(</)(template)(>)",
-      "endCaptures": {
+      "captures": {
```
octref: Shouldn't this still be `endCaptures`?

- `</` — `"punctuation.definition.tag.begin.html"`
- `template`  — `"entity.name.tag.template.html"`
- `>` — `"punctuation.definition.tag.end.html"`

syntaxes/vue-generated.json
```diff
@@ -162,17 +162,9 @@
       ]
     },
     {
-      "begin": "(<)(template)\\b(?=[^>]*lang=('jade'|\"jade\"|'pug'|\"pug\"))(?![^/>]*/>\\s*$)",
-      "beginCaptures": {
```
octref: Why dropping the `beginCaptures`?

syntaxes/vue.yaml
```diff
@@ -42,24 +42,21 @@ patterns:
...
 
   # jade / pug
-  - begin: (<)(template)\b(?=[^>]*lang=('jade'|"jade"|'pug'|"pug"))(?![^/>]*/>\s*$)
-    beginCaptures:
-      '1': { name: punctuation.definition.tag.begin.html }
-      '2': { name: entity.name.tag.template.html }
+  - begin: (<)(template)\b(?=[^>]*lang=[\"'](?:pug|jade)[\"'])(?![^/>]*/>\s*$)
     end: (</)(template)(>)
-    endCaptures:
+    captures:
```
octref: See the comment above — not sure why the change to `captures` from `endCaptures` and dropping `beginCaptures`.

syntaxes/vue-generated.json
Patcher56: As the `captures` keyword is just a shorthand for `beginCaptures` and `endCaptures` and we used the same definitions for `1` and `2` I thought it would be cleaner to just use `captures`.

If you want I can also change this back. The result is the same.

syntaxes/vue-generated.json
Patcher56: Oh, you're right! I will change that asap. Sometimes the cleaner way still is not the best way...

syntaxes/vue-generated.json
octref: I see, just to understand why you made the change.

Yeah I'd prefer them separate. Also with this you don't need to convert the pug/jade part to `(?:)`, although I would be surprised that had any effect since it's enclosed in an `(?=)` lookahead.

octref:
LGTM 👍 Thanks a lot!

PR #1173, opened 2019-04-01
Fix test

octref:


octref:
Just flakey tests from Travis. Maybe time to move off from it...

Issue #1181, opened 2019-04-03
Linting problems originating from vetur are not cleared upon closing a file


jlave-dev:
- [x] I have searched through existing issues
- [x] I have read through [docs](https://vuejs.github.io/vetur)
- [x] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

## Info

- Platform: macOS
- Vetur version: 0.18.0
- VS Code version: 1.32.3

## Problem

Linting problems originating from vetur are not cleared upon closing a file.

## Reproducible Case

* Install ESLint and Vetur extensions for VS Code. Ensure settings.json has the following:
```
"eslint.validate": [
        {
            "language": "vue"
        }
]
```

* Clone [veturpack](https://github.com/octref/veturpack) and open it within VS Code.
* Open a js file and delete a parenthesis. An error is shown in the "Problems" panel.

![](https://i.imgur.com/OZ4U85c.png)

Upon saving and closing the file, the error disappears from the panel.

* Open Counter.vue and delete a curly brace from within the `script` block. Several errors are shown in the "Problems" panel. Some originate from eslint while others originate from Vetur.

![](https://i.imgur.com/Z5760R6.png)

Upon saving and closing the file, eslint errors disappear from the panel. However, Vetur errors remain. I expect all errors to disappear when the file is closed.

![](https://i.imgur.com/U7lsaYF.png)

octref:
Closed through #1182.

PR #1182, opened 2019-04-04
Clear diagnostics on document close

jlave-dev:
Addresses https://github.com/vuejs/vetur/issues/1181. Based on [vscode-eslint's handling](https://github.com/Microsoft/vscode-eslint/blob/ec619b2baa8e7500a451df4156f1973acce315df/server/src/eslintServer.ts#L699) of diagnostics when a document is closed.


octref:
That's correct according to the spec: https://microsoft.github.io/language-server-protocol/specification#textDocument_publishDiagnostics.

Thanks!

Issue #810, opened 2018-06-06
Add scaffold for Typescript & vue-class-component / vue-property-decorator


ascott18:
This was briefly mentioned in https://github.com/vuejs/vetur/issues/669, but was not the primary point of that issue. 

It would be nice to get a scaffold for using vue-class-component and vue-property-decorator. Something like

``` html
<script lang="ts">
import { Vue, Component, Prop, Watch } from 'vue-property-decorator';

@Component({
 
})
export default class extends Vue {
  
}
</script>
```

I included Prop and Watch as imports in this example as they're likely to be used quite commonly. The other property decorators are less likely and can always be imported as needed.

octref:
I'll do #1151 instead.

Issue #1151, opened 2019-03-14
Customizable scaffold snippets


octref:
#903 is not a good solution. As time goes on we'll have more and more snippet coming in, and we can't just include them all. We should provide a way to let users customize their scaffold snippets.

coyotte508:
It would be cool to be able to customize "script wtih typescript" snippet to use something like `vue-property-decorator`.

marius815:
I completely agree with being able to customize your own scaffold snippets, as I feel even the above PR is lacking *my* features.

sandalwoodsh:
Agreed. Allowing users the ability to customize their snippets would be awesome.

jaiko86:
I am all for the customizable scaffolding.

At my office, we use the following:
```
<template>
</template>
<script>
</script>
<style lang="scss" scoped>
@import 'global.scss`;
</style>
```
I'd like to be able to do this automatically every time I make a new `.vue` file.

coyotte508:
Would be nice if there was a way to donate money to specific issues :)

octref:
This is on the top of my todo list, but just sayin', I wouldn't mind a cup of coffee 😉
https://issuehunt.io/r/vuejs/vetur/issues/1151

octref:
Here's the rough plan:

- (User level config) Create a folder under `snippets` of VS Code's user data dir:
  - Windows: `%APPDATA%\Code\User\snippets\vetur`
  - macOS: `$HOME/Library/Application Support/Code/User/snippets/vetur`
  - Linux: `$HOME/.config/Code/User/snippets/vetur`
- (Project level config) Create a folder under `.vscode` in the project root, `.vscode/vetur/snippets/`.
- Project level snippets always show up **before** user level snippets.
- Allow `.vue` components in the `snippets` folder. For example, `snippets/foo.vue` should show a completion item `Scaffold with foo.vue`.
- You can have top-level folders named after block names and custom blocks. For example, `snippets/style/my-style.vue` should show a completion `style with my-style`.
- You can use [Textmate snippet syntax](https://code.visualstudio.com/docs/editor/userdefinedsnippets) in the `.vue` files.

IssueHuntBot:
@khoanguyen96 has funded $20.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/72314479/issues/1151) to receive this reward.
[...]

PR #1198, opened 2019-04-11
Typescript Scaffold

juliankrieger:
Added a typescript scaffold as issued in #810. I'm open for any recommendations on what to do better.


octref:
I don't think I can ever cover all scaffolds so I'll do #1151 instead. Sorry.

PR #1207, opened 2019-04-16
Fix invalid diagnostic range when template block is not placed on the top of file

ktsn:
Template diagnostics will be placed at invalid positions when the `<template>` block is not on the top of the file. I've fixed how we add `<template>` tag when we pass raw template code to vue-eslint-parser as same as [existing code](https://github.com/vuejs/vetur/blob/master/server/src/modes/template/services/htmlValidation.ts#L24).


octref:
Thanks!

PR #1208, opened 2019-04-17
Support v-if type narrowing

### LINKED ISSUE #1204: Support type narrowing by v-if on template diagnostic
Like TypeScript's `if` statement, `v-if`, `v-else-if` and `v-else` should narrow union types which are used in their conditions.

```vue
<template>
  <p v-if="!messages">Loading...</p>
  <div v-else> <!-- messages should be narrowed to string[] here -->
    <p v-for="m in messages" :key="m">
      {{ m }}
    </p>
  </div>
</template>

<script lang="ts">
import Vue from 'vue'

export default Vue.extend({
  data() {
    messages: null as string[] | null
  }
})
</script>
```

ktsn:
fix #1204

### DIFF
--- a/server/src/services/typescriptService/transformTemplate.ts
+++ b/server/src/services/typescriptService/transformTemplate.ts
@@ -20,6 +20,8 @@ const globalScope = (
 
 const vOnScope = ['$event', 'arguments'];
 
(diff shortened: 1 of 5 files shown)


octref:
Very nice work. Thanks a lot!

PR #1209, opened 2019-04-17
Support filter diagnostics

### LINKED ISSUE #1206: Support filter on template diagnostic
It recognizes filters as bitwise OR (`|`) currently. Since we cannot detect a filter function type from component, it would be always treated as `Function` type.

ktsn:
fix #1206


octref:
Thanks! BTW it's `existence` not `existance` 😅 
Since we don't have a way to see the globally registered filter type let's just ignore it for now. Same for mixins...Maybe one day we can resolve them.

PR #1210, opened 2019-04-18
Support scoped slot diagnostics

### LINKED ISSUE #1203: Support scoped slot on template diagnostic
It currently just ignores `slot-scope` and `v-slot`. We need to add declared variables by scoped slot in local scope.

```vue
<template>
  <MyComp v-slot="{ foo }"> <!-- foo will be in local scope -->
    <p>{{ foo }}</p> <!-- should not provide error here -->
  </MyComp>
</template>
```

ktsn:
fix #1203 

Should be merged after #1208

### DIFF
--- a/test/interpolation/diagnostics/basic.test.ts
+++ b/test/interpolation/diagnostics/basic.test.ts
@@ -83,7 +83,8 @@ describe('Should find template-diagnostics in <template> region', () => {
           message: "Argument of type 'string' is not assignable"
         }
       ],
-      loose: true
+      // eslint-plugin-vue generates diagnostics too
+      skipSameDiagnosticCountAssert: true
     },
     {
       file: 'v-slot.vue',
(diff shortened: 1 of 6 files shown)


octref:
Merged #1208. Please try to rebase or merge master. Thanks!

ktsn:
Rebased!

octref:
Retrigger CI.

test/interpolation/diagnostics/basic.test.ts
```diff
@@ -211,4 +222,5 @@ describe('Should find template-diagnostics in <template> region', () => {
 interface TemplateDiagnosticTest {
   file: string;
   diagnostics: vscode.Diagnostic[];
+  loose?: boolean;
```
octref: This is ambiguous because initially I thought it means that the expected message could be `abc` to match `abcdef`, instead of comparing the lengths of the diagnostics count.

I renamed it to `assertSameDiagnosticCount`.

Actually, why do we need to introduce this setting?

ktsn: Because I want to test that 1) scoped slot props are correctly handled (`a` and `b` should not provide error) and 2) type checking inside scoped slot works as same as the others (`c` should provide error). If we don't check actual and expected diagnostics count, 1) will always pass.

I need loose version assertion on `v-if-and-v-for.vue` because it also provides eslint error.

octref: Thanks for the explanation! I just made the naming a bit more explicit.

PR #1222, opened 2019-04-24
Avoid virtual template code is broken when attribute name has hyphen

ktsn:
Currently, when a template includes some attribute having hyphen on its name, the virtual Vue template will be broken and provides numerous errors. This is because the transformer try to generate the attribute name as identifier even though hyphen is invalid character as identifier.

I've simply changed them to string literal to accept hyphen.


ktsn:
Oh, CI is failing. I'll investigate it.

ktsn:
That's weird. Auto complete test passes on my local 🤔

octref:
When that many tests fail it's probably because VLS crashed. All tests passing on my local machine too though so I'm merging this.

PR #1228, opened 2019-04-24
Handle v-on value correctly when it is single expression

### LINKED ISSUE #1227: Vetur fail to start => vOnExp.body.map is not a function

## Info

- Platform: Win
- Vetur version: 0.19.0
- VS Code version: 1.33.1

## Problem

Vetur fail to start. Output: https://justpaste.it/6n1ik

ktsn:
fix #1227 

I'd been casting v-on directive value to `AST.VOnExpression` but it actually can be `ESLintExpression` when the value is 1) simple path `@click="foo.bar"` or 2) function expression `@click="() => foo = 123"`.

ref: https://github.com/mysticatea/vue-eslint-parser/blob/0f241279dc412ff4bb0faa9211233a944465f9b4/src/script/index.ts#L816

In the case 1), there was no issues found because it actually ignores the expression and no diagnostic was provided.

The case 2) causes the issue #1227.


octref:
LGTM, thanks!

PR #1244, opened 2019-04-25
Fix template literal breaks virtual template file

ktsn:
fix #1230

I'm not sure the cause but looks like ts printer emits broken template literal when we reuse its AST node for print. Recreating them solves the issue.

PR #1245, opened 2019-04-25
Fix object literal key breaks virtual template file

### LINKED ISSUE #1237: Veture 0.19.0 throwing multiple error messages.

## Info

- Platform: macOs
- Vetur version: 0.19.0
- VS Code version: 1.33.1

## Problem

In code edit, error (red squiggle) on word 'template'

Seeing multiple error messages stating can't find xxx, as example:

```
Cannot find name 'drag'.
Cannot find global type 'BigInt'.
Expected 3 arguments, but got 7.
The left-hand side of an arithmetic operation must be of type 'any', 'number', 'bigint' or an enum type.
```
Have also seen message that my .vue file is not a module.

Changing 
```
<template>
  <form v-on:submit.prevent class="toolbar grid5 drag-disabled">
    <slot></slot>
  </form>
</template>
```
to 
```
<template lang="html">
  <form v-on:submit.prevent class="toolbar grid5 drag-disabled">
    <slot></slot>
  </form>
</template>
```
seems to solve the problem, at least not reporting any errors.

Here's my .vue file

```
<template lang="html">
  <form v-on:submit.prevent class="toolbar grid5 drag-disabled">
    <slot></slot>
  </form>
</template>

<script>
export default {

}
</script>

<style lang="less" scoped>
form.toolbar.grid5 {
  display: grid;
  grid-template-columns: repeat(5, 1fr);
  grid-gap: 0.3em 0.6em;
  grid-auto-flow: dense;
  align-items: center;
}

form.toolbar {
  padding: 5px;
  background-color: @soft-black;
  border-radius: 4px;
  position: absolute;
  width: calc(100% - 30px);
  bottom: 10px;
}
</style>
```

ktsn:
Possibly fix... #1237 

This is the same kind of issue with #1244. String and Numeric literal in object literal property key emits invalid characters when reusing AST.

### DIFF
--- a/server/src/services/typescriptService/transformTemplate.ts
+++ b/server/src/services/typescriptService/transformTemplate.ts
@@ -441,9 +441,16 @@ export function getTemplateTransformFunctions(ts: T_TypeScript) {
   ): ts.ObjectLiteralElementLike {
     let res;
     if (ts.isPropertyAssignment(el)) {
-      const name = !ts.isComputedPropertyName(el.name)
-        ? el.name
-        : ts.createComputedPropertyName(injectThis(el.name.expression, scope, start));
+      let name: ts.PropertyName;
+      if (ts.isComputedPropertyName(el.name)) {
+        name = ts.createComputedPropertyName(injectThis(el.name.expression, scope, start));
+      } else if (ts.isStringLiteral(el.name)) {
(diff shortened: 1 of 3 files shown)


octref:
Thanks!

PR #1255, opened 2019-04-26
Support external script for template interpolation

### LINKED ISSUE #1254: SFC with external src file errors with Template Interpolation Service enabled
- [X] I have searched through existing issues
- [X] I have read through [docs](https://vuejs.github.io/vetur)
- [X] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

## Info

- Platform: Windows 10
- Vetur version: Vetur 0.19.2
- VS Code version: 
Version: 1.33.1 (system setup)
Commit: 51b0b28134d51361cf996d2f0a1c698247aeabd8
Date: 2019-04-11T08:27:14.102Z
Electron: 3.1.6
Chrome: 66.0.3359.181
Node.js: 10.2.0
V8: 6.6.346.32
OS: Windows_NT x64 10.0.17763

## Problem
Vetur generates an error on a SFC template when the TS script section is in an external file.  This is currently a very common use case for existing TS projects as TS support in SFC was limited until now.  
```
File 'c:/XXXXXX/components/EventDetailsHeader.vue' is not a module.
```
My script section is as follows:
``` javascript
<script lang="ts" src="./EventDetailsHeader.ts"></script>
```

If I move the code from the external file directly to the script section I see the TS Language server error, but after restarting VS Code the error is no longer reported, and the new type features appear to be working correctly

## Reproducible Case

1. In a TS project (Vue CLI 3)
2. Add a new MyComponent.vue component with
``` javascript
<template>
  
</template>

<script lang="ts" src="./MyComponent.ts"/>
```
3. Add a new MyComponent.ts typescript file with
``` javascript
import Vue from "vue";

export default Vue.extend({
  props: {
    
  },
  data() {
   return {
     
   };
  },
  methods: {
    
  }
});
```

If needed I can post a github repo, I'm just tied up at the moment.

ktsn:
fix #1254

### DIFF
--- a/server/src/embeddedSupport/vueDocumentRegionParser.ts
+++ b/server/src/embeddedSupport/vueDocumentRegionParser.ts
@@ -68,7 +68,7 @@ export function parseVueDocumentRegions(document: TextDocument) {
           if (lastAttributeName === 'src' && lastTagName.toLowerCase() === 'script') {
             let value = scanner.getTokenText();
             if (value[0] === "'" || value[0] === '"') {
-              value = value.substr(1, value.length - 1);
+              value = value.slice(1, value.length - 1);
             }
             importedScripts.push(value);
           }
--- a/server/src/embeddedSupport/test/embeddedSupport.test.ts
+++ b/server/src/embeddedSupport/test/embeddedSupport.test.ts
@@ -6,7 +6,7 @@ import { getSingleLanguageDocument, getSingleTypeDocument, getLanguageRangesOfTy
 suite('New Embedded Support', () => {
   const src = `
 <template>
- 
+
 </template>
 <script>
 export default {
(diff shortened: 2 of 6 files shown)


outofmemoryagain:
Thanks for the great work on these new features, I"m really excited about the direction Vetur is moving.  I'm really trying to advocate TS + Vue and this will be a big help :-)

server/src/embeddedSupport/vueDocumentRegionParser.ts
```diff
@@ -68,7 +68,7 @@ export function parseVueDocumentRegions(document: TextDocument) {
           if (lastAttributeName === 'src' && lastTagName.toLowerCase() === 'script') {
             let value = scanner.getTokenText();
             if (value[0] === "'" || value[0] === '"') {
-              value = value.substr(1, value.length - 1);
```
octref: What's the reason for this change?

ktsn: The previous code didn't actually extract `src` value correctly because `String#substr`'s 2nd argument is `length` - the result includes ending `"` if we pass `value.length - 1` to there.
On the other hand `String#slice`'s 2nd argument is `endIndex` which the code seems to intend here.

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice

octref: I see, thanks. Also CI is green now.

PR #1267, opened 2019-05-02
Support statement expression in arrow function block

ktsn:
By this PR, interpolation service supports following usage:

```vue
<template>
  <button @click="() => { foo + 1 }">Increment</button>
</template>
```

I only transform `StatementExpression` in arrow function block because no one would use other statements in template...


octref:
LGTM, thanks!

PR #1273, opened 2019-05-03
fix: Re-add Quasar Framework pre v1 support

rstoenescu:
TLDR; **I added support for the `vetur` Object in package.json (which specifies the location of Vetur's needed tags and attributes jsons) specifically for Quasar. A later change in Vetur code made this feature available for all frameworks, but removed Quasar support as a result, which is ironic.** Quasar developers are missing the autocompletion feature since 5 November.

### Reasoning

This PR re-adds support for Quasar Framework (< v1) when using Quasar CLI.

I initially introduced the concept of using package.json > vetur Object to specify the location of vetur tags and attributes file with [PR](https://github.com/vuejs/vetur/commit/038fe3774cafbdaeda6b6877e1bff21deeb8619d#diff-edc9333e4c2f452171b45de5fcbe9376R28). This was done for Quasar Framework only.

Later on, this concept became the default way of dealing with any frameworks through [PR](https://github.com/vuejs/vetur/commit/d42b662762ed48f10ff2a1528759e5abc41a7b00#diff-f0b24377167be7ffee9c03f0072ba2fd) by @Zenser . But while introducing this, Quasar became undetectable, so its support dropped.

In a Quasar CLI project folder, Quasar Framework pre v1 (eg: v0.17) is supplied by `quasar-cli` package which is specified in package.json > devDependencies through `quasar-framework` package (`quasar-framework` is a dependency of `quasar-cli`, not of the project itself). This was missed by @Zenser 's PR, because Vetur ended up checking quasar-cli's package.json for the `vetur` Object that I introduced -- which is not there. My initial code was detecting `quasar-cli` package and then it was checking `node_modules/quasar-framework/package.json` to get the details. My initial PR added a test-case specifically for Quasar, but it was directly using `quasar-framework` package as a dependency (as in Vue CLI usage of a project with Quasar), instead of `quasar-cli`. So the tests were successful -- since `quasar-framework` dependency was specified in the test project folder, which is the NPM package that has the vetur Object in its package.json.

The current PR detects if `quasar-cli` is specified in a project's devDependencies and virtually adds `quasar-framework` to the in-memory package.json > dependencies. This allows Vetur to correctly get the provider a few lines below this in the `for(dep in dependencies)`.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -6,7 +6,7 @@
 
 Vue tooling for VS Code, powered by [vue-language-server](https://github.com/vuejs/vetur/tree/master/server).
 
-Try it out with [Veturpack](https://github.com/octref/veturpack)!  
+Try it out with [Veturpack](https://github.com/octref/veturpack)!
 
 🎉 VueConf 2017 [Slide](https://www.dropbox.com/sh/eb4w8k3orh0j391/AAB3HaJexbGLa2tCP14BI8oJa?dl=0) & [Video](https://www.youtube.com/watch?v=05tNXJ-Kric) 🎉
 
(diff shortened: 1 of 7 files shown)


octref:
LGTM. Sorry for the situation and thanks for the detailed explanation!

rstoenescu:
@octref Thanks for merging.

Will write a doc page for Quasar recommending an IDE to use (VSCode) and what extensions to install for it -- and this will include Vetur.

PR #1287, opened 2019-05-08
Infer $event type more accurately

### LINKED ISSUE #1281: Template Interpolation Service - @onClick MouseEvent
When using the new Template Interpolation Service with typescript a `@click=onHandleClick($event)` will get the error `Argument of type 'Event' is not assignable to parameter of type 'MouseEvent'.` When using the following handler

```
private onHandleClick (event: MouseEvent) { }
```

I believe this all comes down to this line here. Where the vOn transform assumes everything is of type `Event`. 
https://github.com/vuejs/vetur/blob/master/server/src/services/typescriptService/transformTemplate.ts#L196

Is there anyway to make this more specific? Check for all mouse related listeners and swap over to `MouseEvent`?

Not sure what the best workaround is at this point. I could do something like this
```
private onHandleClick (event: Event) {
    const mouseEvent = event as MouseEvent;
}
```
But this seems like a waste to add this extra variable assignment just to work around this issue. Plus it just feels wrong to have a handler with an incorrectly typed parameter.

Any thoughts are appreciated :) and I have no problem sending in a PR if its useful but I am guessing it's easier to just provide information at this point.

ktsn:
fix #1281 

Utilizing `HTMLElementEventMap` which has all possible event name and event type pairs for `$event` type.

Note that `ComponentData` bridge type is only meant to be used for inferring `$event` type currently. So it will skip type check in case of passing method name directly and `v-on` without argument like below:

```vue
<template>
  <div>
    <!-- Error: mismatching onClick arg type and $event type -->
    <input @keydown="onClick($event)" />
    
    <!-- NOT Error: it does not check implicit argument type -->
    <input @keydown="onClick" />
    
    <!-- NOT Error -->
    <input v-on="{ keydown: onClick }" />
  </div>
</template>

<script lang="ts">
import Vue from 'vue'

export default Vue.extend({
  methods: {
    onClick(event: MouseEvent): void {}
  }
})
</script>
```

This is because there are no guarantee that the event is from native DOM elements. e.g. when some Vue component emits string value via `change` event which is the same name with native DOM event but not value type.

On the other hand, `$event` will only be used in native event handlers. So we can safely infer its type from `HTMLElementEventMap`.

In the future, we would be able to analyze it more accurate if we can detect which tag in template is component.

### DIFF
--- a/server/src/services/typescriptService/bridge.ts
+++ b/server/src/services/typescriptService/bridge.ts
@@ -14,7 +14,7 @@ type ComponentListeners<T> = {
 interface ComponentData<T> {
   props: Record<string, any>;
   on: ComponentListeners<T>;
-  directives: any[]
+  directives: any[];
 }
 export declare const ${renderHelperName}: {
   <T>(Component: (new (...args: any[]) => T), fn: (this: T) => any): any;
--- a/server/src/services/typescriptService/bridge.ts
+++ b/server/src/services/typescriptService/bridge.ts
@@ -1,4 +1,4 @@
-import { renderHelperName, componentHelperName, iterationHelperName, listenerHelperName } from './transformTemplate';
+import { renderHelperName, componentHelperName, iterationHelperName } from './transformTemplate';
 
 // This bridge file will be injected into TypeScript language service
 // it enable type checking and completion, yet still preserve precise option type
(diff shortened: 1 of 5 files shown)


octref:
LGTM — thanks!

PR #1310, opened 2019-05-23
Relax $event type to any

### LINKED ISSUE #1306: The $event is wrong when custom event

## Info

- Platform: Linux
- Vetur version: 0.21.0
- VS Code version: 1.33.1

## Problem

When use custom event the `$event` is wrong because this can be a value emited
 and not `Event`
Screenshot_20190523_134404

The event is emited with :
```ts
private value: Website.FieldValue = ''

@Emit('change')
private onChange () {
  return this.value
}
```

## Reproducible Case

1. Create component
2. Send an event with value, like :
    ```ts
    this.$emit('event', 'value')
    ```
3. Bind the event with inline value like :
    ```vue
    <component @event="onEvent('test', $event)" />
    ```
    The method :
    ```ts
    onEvent (type: string, value: string) {
      console.log(type, value)
    }
    ```

ktsn:
fix #1306 

I said `$event` is only used native event handler but it was not true actually.
We can use it in component event handler and it refers the first parameter of event.
https://vuejs.org/v2/guide/components.html#Emitting-a-Value-With-an-Event

I guess we should relax its type to `any` until we can detect component tag in future.

### DIFF
--- a/server/src/services/typescriptService/bridge.ts
+++ b/server/src/services/typescriptService/bridge.ts
@@ -23,7 +23,7 @@ export declare const ${componentHelperName}: {
   <T>(
     vm: T,
     tag: string,
-    data: ComponentData<HTMLElementEventMap & Record<string, any>> & ThisType<T>,
+    data: ComponentData<Record<string, any>> & ThisType<T>,
     children: any[]
   ): any;
 };


octref:
Retrigger the build

octref:
Thanks! Had some troubles with Travis CI, but should be good now.

PR #1335, opened 2019-06-15
Improve source map position accuracy

ktsn:
Trying to resolve the issue that the source map location is differed when there are extra space / less space like:

```vue
<template>
  <div v-if="
    !loading1
    && !loading2
  ">
    Hi
  </div>
</template>
```

In the above case the expression in `v-if` will be transformed to `!loading && !loading2` (space and line breaks are removed). Then the diagnostics will point wrong location.

To resolve it, I changed the AST transformer to set source map range for every expression node and merge it when generating source map.

With this approach, the diagnostics point correct location but the source map test is failing because it doesn't correct locations other than expression node like `&&` on the above example. I'll investigate how to deal with this problem.

### DIFF
--- a/server/src/services/typescriptService/sourceMap.ts
+++ b/server/src/services/typescriptService/sourceMap.ts
@@ -71,7 +71,7 @@ export function generateSourceMap(
 
   const sourceMapNodes: TemplateSourceMapNode[] = [];
   walkBothNode(syntheticSourceFile, validSourceFile);
-  return sourceMapNodes;
+  return foldSourceMapNodes(sourceMapNodes);
 
   function walkBothNode(syntheticNode: ts.Node, validNode: ts.Node) {
     const validNodeChildren: ts.Node[] = [];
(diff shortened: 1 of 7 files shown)


ktsn:
@octref I guess we should reconsider the way testing sourceMap.
Imagine we have the following expression in template:

```js
123+456
```

The expression will have whitespace between each node after generating valid position AST:

```js
123 + 456
```

In this case the node `123`'s source mapping is like the following:

```
0 -> 0 (`1` -> `1`)
1 -> 1 (`2` -> `2`)
2 -> 2 (`3` -> `3`)
3 -> 3 (`+` -> ` `)
```

The last offset mapping (3 -> 3) seems valid but its locations point the different characters (`+` -> ` `) as a whitespace is inserted in the converted code.

ktsn:
To avoid invalid source mapping by inserting/removing whitespaces, changing `updateOffsetMapping` to only fill intermediate mapping for `Identifier` and some literals which cannot include trivia. It only maps start and end position if it can include any trivia characters.

Example if we have following expression

```
Before:
test  +foo["bar"]

After:
this.test + this.foo["bar"]
```

`test`, `foo` and `"bar"` (identifier, string literal nodes) map entire positions (e.g. 0,1,2,3,4 -> 5,6,7,8,9 for `test`) because they cannot have trivia.

On the other hand `test  +foo["bar"]`, `foo["bar"]` maps only start and end positions (e.g. 0,17 -> 0,27 for `test  +foo["bar"]`) because they can have trivia and we cannot guarantee the correctness other than start / end positions mapping.

After generating mapping for each node, it merges all children mapping so the result will be:

```
0 -> 0: t
1 -> 1: e
2 -> 2: s
3 -> 3: t
4 -> 4:
7 -> 12: f
8 -> 13: o
9 -> 14: o
10 -> 15: [
11 -> 16: "
12 -> 17: b
13 -> 18: a
14 -> 19: r
15 -> 20: "
16 -> 21: ]
17 -> 22:
```

I solved the problem which I mentioned on previous comment by excluding end positions of all merged nodes from mapping test. So this PR is already ready to review / merge.

octref:
Very nice work. Thanks a lot and sorry for the delay!

PR #1338, opened 2019-06-17
substitute `createStatement` with `createExpressionStatement`

fuafa:
`createStatement` is deprecated, use `createExpressionStatement` instead.


octref:
This probably breaks support for TypeScript 2.8. Let me check with CI.

octref:
OK that's fine. Thanks!

PR #1343, opened 2019-06-27
Fix Travis

octref:


DiamondYuan:
ok thanks.That work for my extensions.

PR #1365, opened 2019-07-24
feat: remove vuetify-helper-json and use runtime pkg info

nekosaur:
switched to using `vetur` option inside package.json for auto-completion instead of separate package.

also added support through @nuxtjs/vuetify package

### DIFF
--- a/server/src/modes/template/tagProviders/index.ts
+++ b/server/src/modes/template/tagProviders/index.ts
@@ -7,7 +7,6 @@ import {
   onsenTagProvider,
   bootstrapTagProvider,
   buefyTagProvider,
-  vuetifyTagProvider,
   gridsomeTagProvider,
   getRuntimeTagProvider
 } from './externalTagProviders';
@@ -27,7 +26,6 @@ export let allTagProviders: IHTMLTagProvider[] = [
   onsenTagProvider,
   bootstrapTagProvider,
   buefyTagProvider,
-  vuetifyTagProvider,
   gridsomeTagProvider
 ];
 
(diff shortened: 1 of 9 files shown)


server/src/modes/template/tagProviders/index.ts
```diff
@@ -76,9 +76,12 @@ export function getTagProviderSettings(workspacePath: string | null | undefined)
     if (dependencies['buefy']) {
       settings['buefy'] = true;
     }
-    if (dependencies['vuetify']) {
+    if (dependencies['vuetify'] || devDependencies['vuetify']) {
       settings['vuetify'] = true;
     }
+    if (dependencies['@nuxtjs/vuetify'] || devDependencies['@nuxtjs/vuetify']) {
+      dependencies['vuetify'] = true;
```
kevinmarrec: Shouldn't it be :
```suggestion
      settings['vuetify'] = true;
```
?

nekosaur: No. Maybe the nuxt check should happen _before_ the vuetify check so that `settings['vuetify']` gets set as well 🤷‍♂ . I am not actually 100% sure what setting the `settings` object accomplishes 😕 . 

The actual loading of the autocomplete data specified in `vetur` package.json option happens here https://github.com/vuejs/vetur/blob/4056e3d3c9c49737f1ecb17383bb8b8340089b7e/server/src/modes/template/tagProviders/index.ts#L116 and that depends on what is inside the `dependencies` object. Since vuetify is not a direct dependency when using nuxtjs module, we fake it by inserting vuetify.

kevinmarrec: Good catch, then all good I believe

octref:
LGTM, thanks!

PR #1371, opened 2019-07-30
Avoid template diagnostic error on empty v-on

ktsn:
Sometimes the users want to write `v-on` with empty value like `@click.stop`. In this case, Vetur produce an error because it passes `true` as the value on virtual source file.

Screenshot 2019-07-30 at 10 18 23 AM

Changing the value to noop solves this issue.


octref:
Thanks!

PR #1382, opened 2019-08-11
fix prettyhtml options always be covered by default

alltennis:


octref:
Please write down what's the issue this fixes, thanks.

Issue #1433, opened 2019-09-16
Formatting for sass


TheRealSyler:
Im currently working on making my sass formatter work with this extension, is there a way to get the `TextDocument` Object from vscode ?
since i made my formatter in a vscode extension i worked with the `TextDocument` from vscode, right now im looping trough the text two times one time to create an object that is similar to the `TextDocument` object and one time to format the code, i could probably rewrite the formatter but i don't want to spend my time doing that since it already works.

i could just add formatting for vue files in my extension, but i think it would be nice to have it in this extension, so that the vetur supports everthing to do with vue and the sass extension just supports sass files, i also added hover info for some properties, and im working on a linter all stuff that can also be added to this extension.

octref:
`TextDocument` in VS Code is different from `TextDocument` in LSP. What are you missing from the LSP `TextDocument`?

Vue Language Server works with multiple modes. See https://github.com/vuejs/vetur/tree/master/server/src/modes
The way this should be added is `sass` added as one more mode along `style` (which is really a misnomer, it should be css-less-scss).

If you install the dependencies in `server`, this signature would give you all info needed for implementing formatting: https://github.com/vuejs/vetur/blob/master/server/src/embeddedSupport/languageModes.ts#L64

Thanks for your work on SASS! It has been a requested many times for Vetur to add better sass support.

TheRealSyler:
@octref  my [fork ](https://github.com/TheRealSyler/vetur/blob/master/server/src/embeddedSupport/languageModes.ts#L145)i already have added a new mode.
 
the only problem is that i need a document object that has `lineAt` which needs to return a line object that has a `text` and `lineNumber` property, since the lSP `TextDocument` doesn't have the properties that i need i was wondering if its possible to get the vscode version of the `TextDocument`.

[this](https://github.com/TheRealSyler/sass-formatter/blob/master/src/format.provider.ts#L57) is the class that i currently use to convert text to an object similar to vscodes `TextDocument`, creating this class can be skipped, as you can see [here](https://github.com/TheRealSyler/vscode-sass-indented/blob/master/src/format/format.provider.ts#L53) i just pass the vscode `TextDocument`, in the [vetur implementation](https://github.com/TheRealSyler/vetur/blob/master/server/src/embeddedSupport/sassLanguageMode.ts#L54) i need to create the 'SassTextDocument' class to make it work.

octref:
Fixed by #1658.

PR #1434, opened 2019-09-18
Sass Formatter

TheRealSyler:
Related to #1433

not sure if the return values of the unused functions in the `SassLanguageMode` are an issue.

### DIFF
--- a/server/src/modes/style/sass/sassLanguageMode.ts
+++ b/server/src/modes/style/sass/sassLanguageMode.ts
@@ -6,7 +6,7 @@ import { Position, TextEdit } from 'vscode-css-languageservice';
 
 import { DocumentContext } from '../../../types';
 
-import { SassFormatter, SassTextDocument } from 'sass-formatter';
+import { SassFormatter, SassTextDocument, SassFormatterConfig } from 'sass-formatter';
 
 export class SassLanguageMode implements LanguageMode {
   private config: any = {};
@@ -40,7 +40,7 @@ export class SassLanguageMode implements LanguageMode {
     return [];
   }
   format(document: TextDocument, range: Range, formattingOptions: FormattingOptions) {
-    let sassConfig: any = {
+    const sassConfig: SassFormatterConfig = {
       convert: true,
       deleteCompact: true,
       deleteEmptyRows: true,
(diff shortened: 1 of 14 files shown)


server/src/modes/style/sass/sassLanguageMode.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+  format(document: TextDocument, range: Range, formattingOptions: FormattingOptions) {
+    let sassConfig: any = {
+      convert: true,
+      deleteCompact: true,
+      deleteEmptyRows: true,
+      deleteWhitespace: true,
+      replaceSpacesOrTabs: true,
+      setPropertySpace: true
+    };
```
octref: Can you extract this into an interface exported from this file? We might have a default config at some point.

server/src/modes/style/sass/sassLanguageMode.ts
```diff
@@ -0,0 +1,80 @@
...
+    if (this.config.sass && this.config.sass.format) {
+      sassConfig = this.config.sass.format;
```
octref: Use `Object.assign(this.config.sass.format, sassConfig)` to merge them.

octref:
Good work. Thanks a lot!

I have added docs and pushed an integration test. Left some comments as well.

server/src/modes/style/sass/sassLanguageMode.ts
TheRealSyler: i already made the interface [here](https://github.com/TheRealSyler/sass-formatter/blob/master/src/format.provider.ts#L32), i can just import the it from `sass-formatter`, maintaining two instances of the same interface seems unnecessary.

TheRealSyler:
i couldn't test it locally because i installed linux, and i get an error when i try to start the debug server, but this should address your comments, and in theory it should work without braking anything, but every time i think this should work something brakes.

server/src/modes/style/sass/sassLanguageMode.ts
TheRealSyler: added in the newest commit.

octref:
Retriggering CI

TheRealSyler:
moved to #1658

Issue #1427, opened 2019-09-11
Allow applying `templateInterpolationService` only to `lang="ts"` components


jackkoppa:
- [x] I have searched through existing issues
- [x] I have read through [docs](https://vuejs.github.io/vetur)
- [x] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

## Info

- Platform: macOS
- Vetur version: 0.22.2
- VS Code version: 1.38.0

### Feature Request - add a third option for `templateInterpolationService`, `ts-only`

As mentioned in #1246, the `templateInterpolationService` works well in a lot of cases, but will quite regularly throw errors for JavaScript components, because without proper attention paid to adding type annotations to a component using `lang="ts"`, it's very unlikely that a JS Vue component will be able to be parsed correctly.

So, in JS components, there are a lot of errors that end up looking like this, when setting `"vetur.experimental.templateInterpolationService": true`

![Screen Shot 2019-09-11 at 7 37 39 AM](https://user-images.githubusercontent.com/14101491/64694472-b979d480-d467-11e9-80b8-73ef72b78199.png)

In our case, and potentially other repos that are attempting to get TS benefits in a codebase that was originally JS-only, some components are TS, while many remain JS. 

It seems likely that only TS components will see useful benefits from `templateInterpolationService`, 
**so the proposal is to add a third option to the settings, `ts-only`**

Changing [these lines](https://github.com/vuejs/vetur/blob/master/package.json#L472-L476) to something like 

```json
"vetur.experimental.templateInterpolationService": {
	"type": ["boolean", "string"],
	"default": false,
	"description": "Enable template interpolation service that offers diagnostics / hover / definition / references. Use 'ts-only' to enable only for TypeScript components"
}
```

I'm sure the above can be done much better, hopefully with `enum`, while still maintaining backwards compatibility (default still `false`, still allowing `true` & applying to all files).

I'm very happy to work on this change, if it's an acceptable request & the use case (JS & TS components in the same project, or at least the same workspace) is common enough. I don't have any significant extension development experience, so any other recommendations on how to [structure the config option](https://code.visualstudio.com/api/references/contribution-points#Configuration-property-schema) are welcome. 

## Reproducible Case

(happy to add a modified Veturpack with both JS & TS components, if that's useful)

jackkoppa:
Two notes I forgot:

[...]

PR #1439, opened 2019-09-22
Apply template interpolation validation only to TypeScript components

jackkoppa:
Hi @ktsn - thanks for the direction on this in #1427 

The PR checks if the current document includes `typescript` as a language; if it does not, the template interpolation service will no longer attempt to do validation on that document. As you suggested [here](https://github.com/vuejs/vetur/issues/1427#issuecomment-533380654), though, the other, non-diagnostic features associated with the interpolation service will continue to run, so long as the user has enabled the feature in their settings.

**What I still need assistance with:**
* I'm not certain how we should be checking for `// @ts-check`. Do you happen to know if the `typescript` module provides information about "yes, this file is being type-checked, whether because of its extension or the ts-check directive"? Otherwise, presumably we need to parse the beginning of the script, to find that actual text?
* I wanted to confirm that the method of checking the languages in the current document is OK; after a fair amount of digging, I couldn't find an alternative approach in the codebase

Please absolute suggest any changes or alternative approaches; just wanted to get an initial approach out there. There are currently 2 failing `interpolation` tests; the JSDoc case can certainly be handled if we figure out how to capture the `ts-check` directive

### DIFF
--- a/server/src/modes/template/interpolationMode.ts
+++ b/server/src/modes/template/interpolationMode.ts
@@ -50,7 +50,7 @@ export class VueInterpolationMode implements LanguageMode {
   }
 
   doValidation(document: TextDocument): Diagnostic[] {
-    if (!this.interpolationEnabledInTypeScriptDocument(document)) {
+    if (!this.interpolationEnabled || !this.documentTypeChecked(document)) {
       return [];
     }
 
@@ -218,7 +218,7 @@ export class VueInterpolationMode implements LanguageMode {
     contents: MarkedString[];
     range?: Range;
   } {
-    if (!this.interpolationEnabledInTypeScriptDocument(document)) {
+    if (!this.interpolationEnabled()) {
       return { contents: [] };
     }
 
@@ -257,7 +257,7 @@ export class VueInterpolationMode implements LanguageMode {
   }
 
   findDefinition(document: TextDocument, position: Position): Location[] {
-    if (!this.interpolationEnabledInTypeScriptDocument(document)) {
+    if (!this.interpolationEnabled()) {
       return [];
     }
 
@@ -305,7 +305,7 @@ export class VueInterpolationMode implements LanguageMode {
   }
 
   findReferences(document: TextDocument, position: Position): Location[] {
-    if (!this.interpolationEnabledInTypeScriptDocument(document)) {
+    if (!this.interpolationEnabled()) {
       return [];
     }
 
@@ -356,12 +356,11 @@ export class VueInterpolationMode implements LanguageMode {
 
   dispose() {}
 
-  private interpolationEnabledInTypeScriptDocument(document: TextDocument): boolean {
-    const documentIncludesTypeScript =
-      !!this.vueInfoService && this.vueInfoService.documentIncludesLanguage(document, 'typescript');
-    return (
-      _.get(this.config, ['vetur', 'experimental', 'templateInterpolationService'], true) && documentIncludesTypeScript
-    );
+  private interpolationEnabled(): boolean {
+    return _.get(this.config, ['vetur', 'experimental', 'templateInterpolationService'], true);
+  }
+  private documentTypeChecked(document: TextDocument): boolean {
+    return !!this.vueInfoService && this.vueInfoService.documentIncludesLanguage(document, 'typescript');
   }
 }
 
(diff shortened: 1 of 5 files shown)


jackkoppa:
(should not be merged until we at least address the failing specs, and hopefully the `ts-check` case)

ktsn:
Thank you for the PR! I'll look into the details when I have more extra time. 

> checking for // @ts-check

Looks like we can get this info from `SourceFile#checkJsDirective`.
https://ts-ast-viewer.com/#code/FAehAIAEBcGcFoDGALApog1sRB7AdrNOAIbgC84A5JcEA

We still need to think about how we get source file object in `doValidation`.

> I wanted to confirm that the method of checking the languages in the current document is OK

Looks good to me.

server/src/modes/template/interpolationMode.ts
```diff
@@ -216,6 +221,13 @@ export class VueInterpolationMode implements LanguageMode {
   onDocumentRemoved() {}
 
   dispose() {}
+
+  private interpolationEnabled(): boolean {
+    return _.get(this.config, ['vetur', 'experimental', 'templateInterpolationService'], true);
+  }
+  private documentTypeChecked(document: TextDocument): boolean {
```
jackkoppa: Thanks for the response thus far, @ktsn. Very much appreciated. 

I'll come back to this in the near future, but for now, I'm having trouble accessing `sourceFile.checkJsDirective`; it shows as non-existent when I do the following:

```typescript
  private documentTypeChecked(document: TextDocument): boolean {
    // Very rough changes, just to see if we can get the right AST 
    // without changing providers or services for this class right away

    const { updateCurrentVueTextDocument } = this.serviceHost;
    const { service } = updateCurrentVueTextDocument(document);
    const program = service.getProgram();
    if (!languageServiceIncludesFile(service, document.uri) || !program) {
      return false;
    }
    const fileFsPath = getFileFsPath(document.uri);
    const sourceFile = program.getSourceFile(fileFsPath);
    
    return !!this.vueInfoService && (
      this.vueInfoService.documentIncludesLanguage(document, 'typescript') ||
      // ISSUE: Below, sourceFile, type `SourceFile`, does not contain `checkJsDirective` 
      (sourceFile && (sourceFile as any).checkJsDirective.enabled));
  }
```

jackkoppa: Feel free to copy/paste these changes into the branch to see what I mean. 
TypeScript version seems fine; it's odd to me that the AST viewer includes `checkJsDirective`, but it's not a property available on the `SourceFile` interface.

I'll come back to this when I have time; just leaving the note in case anyone's interested in where I'm currently stuck

ktsn: This looks like not in declaration but available in runtime. Using it with type cast would be OK I think.

F0rsaken:
is this PR dead?

jackkoppa:
Hi! Thanks for asking, @F0rsaken - only dormant because I haven't come back to it yet 😬 

I'm going to dig into this again - I'm struggling a bit with how to get info from the language service about whether it's in a type-checked file, but will report back, hopefully within the next week.

For us, this still remains a bit of a blocker in a few of our repos with mixed JS & TS components, so definitely hoping to get this resolved

F0rsaken:
@jackkoppa same with me. We have project that went from pure JS to JS/TS mix 'cause we wanted transition to TS and this feature would be so helpful! 😄 

I would love to help you with that but I never tried writing VS Code plugins so I dk if I can but maybe I try 😉

jackkoppa:
Note that I pushed some rebasing changes to this branch, but it feels more likely that I'll be opening a new PR on a different branch, to take into account some codebase changes between when I started this PR & now

server/src/modes/template/interpolationMode.ts
jackkoppa: Back to working through approaches to this, for a follow-up PR. But just as a reminder for myself, the reason is that `checkJsDirective` is internal: https://github.com/microsoft/TypeScript/blob/master/src/compiler/types.ts#L3010

It's easier to see this in the AST viewer now, thanks to a change that makes `internal` properties red 🤦‍♂ https://github.com/dsherret/ts-ast-viewer/issues/65

So, looking into additional approaches now...

jackkoppa:
Yay! Finally got to this!

Assuming there will be fixes needed, but an updated & much more ready-to-review PR is here: #1824

PR #1446, opened 2019-10-01
Use typescript info for interpolation completion

### LINKED ISSUE #1129: Template interpolation completion doesn't work with object's props
- [ X ] I have searched through existing issues
- [ X ] I have read through [docs](https://vuejs.github.io/vetur)
- [ X ] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

## Info

- Platform: 
- Vetur version:  0.16.2
- VS Code version: 1.31.1

## Problem
vetur

ktsn:
fix #1129 

To use `ts.LanguageService#getCompletionsAtPosition` to get completion list for interpolation expression. It also combines previous completion info so that the behavior should be backward compatible.

~~There is a known issue when an interpolation expression is invalid. Our source map cannot handle expression position, then provides wrong completion in that case. We need to think how we handle invalid expression in the future.~~
-> Fixed via #1448 

```vue
<template>
  <div>{{ obj. }}</div> <!-- Cannot provide `obj.value` completion -->
</template>

<script lang="ts">
import Vue from 'vue'

export default Vue.extend({
  data() {
    return {
      obj: { value: '' }
    }
  }
})
</script>
```


sabifa:
I would love to see that feature! Could we get a small update about it, please? 😄

octref:
I got completion resolve as well:

image

octref:
Still need to fix the empty completion case `{{ | }}` but this is much better now.

test/interpolation/fixture/completion/Parent.vue
```diff
@@ -35,7 +35,7 @@ export default {
      * My count
      */
     count () {
-      return this.$store.state.count
+      return 42;
```
octref: I had to change this because otherwise the typing would be wrong and cause the whole `this` to have any type...

test/lsp/fixture/client/hover/Basic.vue
```diff
@@ -26,7 +26,7 @@ export default {
   },
   computed: {
     count () {
-      return this.$store.state.count
+      return 42;
```
octref: I had to change this because otherwise the typing would be wrong and cause the whole `this` to have any type...

test/interpolation/fixture/completion/Parent.vue
ktsn: It's probably because of the return type annotation problem... 😢 
https://vuejs.org/v2/guide/typescript.html#Annotating-Return-Types

octref: JSDocs return type doesn't work either. Sad..

PR #1465, opened 2019-10-18
Highlight directives containing numbers

shroudedcode:
Currently directives are not correctly highlighted if their names contain numbers. This makes templates containing "dynamic SVG" pretty unreadable as the names of many [SVG attributes](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute) (`x1`, `k1`, `in2`, etc.) contain numbers. Changing the pattern to allow digits in the directive name fixes this problem:

|Before|After
|---|---
|Before|After

I know that the Vue grammar is technically not meant to encompass SVG, but I don't see how this change negatively affects HTML highlighting.


octref:
Old build failed and is gone. I rebased and added some test.

octref:
Good job. Thank you!

PR #1450, opened 2019-10-07
Fix intellisense and other features when open multiple files

yoyo930021:
Fix #1449

PR #1469, opened 2019-10-22
Fixed an issue with file paths in javascript mode on windows.

Kam1ni:
I work almost exclusively on Windows, and autocomplete barely ever worked. Turns out that the function "languageServiceIncludesFile" checked if the file is inside the project, but used two different types of formatting. The list of filePaths uses "\" and filePath uses "/". Normalizing the path fixed the issue. 

I copied the two variables into chrome for better readability and this was the result before the fix:
image


yoyo930021:
Thanks for @Kam1ni.

I think it's my error on #1450.
I add error path format to scriptFileNameSet.

I send a PR #1472 to fix it.

octref:
This is not root of the problem. This part itself is fine. It's the other parts that's setting wrong paths...

Issue #1449, opened 2019-10-04
Intellisense / type checking stopped working in v0.22.4


Soresu:
- [x] I have searched through existing issues
- [x] I have read through [docs](https://vuejs.github.io/vetur)
- [x] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

## Info

- Platform: Windows 10 x64
- Vetur version: 0.22.4
- VS Code version: 1.38.1
- Already reinstalled Visual Studio Code, Vetur
- Happens on multiple PC
- Tried to remove other plugins

## Problem

Intellisense stopped working properly in Vue files. Although it works in the first opened vue file or if it was opened in the previous session.
Downgrading to Vetur 0.22.3 fixed the issue.

## Reproducible Case

1. Close all tabs in the current project
2. Close and open Visual Studio Code
3. Open a vue file
4. Wait until the plugins are loaded
5. Hovering over/Ctrl+space at imported objects, variables doesn't show any info
6. Close and open Visual Studio Code
7. Intellisense works in the previously opened file

borradaniele:
Same here, linking between files does not work with cmd and intellisense does not shows up, already tried to uninstall and reinstall.
Seems like occasionally it starts working again for some times and then stops again, not sure about the events chain that causes this tho.

Platform: Mac OS Mojave 10.14.6
VS Code version: 1.38.1
Vetur version: 0.22.4

If you need more info or log let me know.

alexsasharegan:
Same on Linux.

Vetur: 0.22.4

**VSCode Stats:**
Version: 1.38.1
Commit: b37e54c98e1a74ba89e03073e5a3761284e3ffb0
Date: 2019-09-11T13:30:08.229Z
Electron: 4.2.10
Chrome: 69.0.3497.128
Node.js: 10.11.0
V8: 6.9.427.31-electron.0
OS: Linux x64 5.0.0-31-generic

yoyo930021:
https://github.com/vuejs/vetur/issues/1317#issuecomment-537488062

Sheng-Long:
Same issue for me, although it always seems to work in at least one file, but not the others

huangysss:
+1

hkk2018:
+1

matthew-ch:
Similar issue for my project using typescript.

VSCode version: 1.39.0
OS: macOS Catalina 10.15
Vetur version: 0.22.4

Downgrading to 0.22.3 works.

[...]

PR #1472, opened 2019-10-22
Fix add to scriptFileNameSet path format

yoyo930021:
Fixed #1450 #1449

### DIFF
--- a/server/src/services/typescriptService/serviceHost.ts
+++ b/server/src/services/typescriptService/serviceHost.ts
@@ -139,7 +139,7 @@ export function getServiceHost(
 
     if (isVirtualVueTemplateFile(fileFsPath)) {
       localScriptRegionDocuments.set(fileFsPath, doc);
-      scriptFileNameSet.add(fileFsPath);
+      scriptFileNameSet.add(filePath);
       versions.set(fileFsPath, (versions.get(fileFsPath) || 0) + 1);
     }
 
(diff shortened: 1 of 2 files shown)


yoyo930021:
@octref
Sorry I have an error fixed in #1450.
Path format was error in Typescript Program when using windows.

This PR can fix it.

server/src/services/typescriptService/serviceHost.ts
```diff
@@ -289,7 +289,7 @@ export function getServiceHost(
                 TextDocument.create(uri.toString(), 'vue', 0, tsModule.sys.readFile(resolvedFileName) || '')
               );
               localScriptRegionDocuments.set(resolvedFileName, doc);
-              scriptFileNameSet.add(resolvedFileName);
+              scriptFileNameSet.add(uri.path);
```
octref: Same as above

server/src/services/typescriptService/serviceHost.ts
```diff
@@ -219,7 +219,7 @@ export function getServiceHost(
               TextDocument.create(uri.toString(), 'vue', 0, tsModule.sys.readFile(fileName) || '')
             );
             localScriptRegionDocuments.set(fileName, doc);
-            scriptFileNameSet.add(fileName);
+            scriptFileNameSet.add(uri.path);
```
octref: You can use getFilePath from `paths.ts` util.

octref:
Thanks, there are some documentations in `paths.ts` that you can read to deal with Windows path handling.

server/src/services/typescriptService/serviceHost.ts
yoyo930021: localScriptRegionDocuments set uri.fsPath
scriptFileNameSet set resolvedFileName
Path format is be right.

server/src/services/typescriptService/serviceHost.ts
yoyo930021: I changed the method.

octref:
I made the path naming consistent and left some more notes in `paths.ts`. I tested and this works now on Windows.

PR #1505, opened 2019-11-16
Update the Quasar Framework's URL in framework.md

Juners:
Seems like the URL has changed, as it can be seen in their GitHub repository.

[The new URL](https://quasar.dev/)


octref:
Thanks!

PR #1567, opened 2019-11-28
feat: support prettier-stylelint

### LINKED ISSUE #712: Prettier-Stylelint Support
Would It be feasible to add support for [prettier-stylelint](https://github.com/hugomrdias/prettier-stylelint) like [prettier-eslint](https://github.com/prettier/prettier-eslint)?

I love working on VSCode using [prettier-vscode](https://github.com/prettier/prettier-vscode#prettierdisablelanguages-default-) and not so long ago they added support for prettier-stylint, but that only works for css and related files. So I would like to have the same facility when working on .vue files.

I could start working on it with some guidance if It is considered as a good addition to Vetur
I already started reading through the source code to have some understanding of how it could be done.

zyy7259:
close #712


octref:
Please reopen when you have time to address the feedback. I'm just closing so I know I don't need to look at it right now. Thank you!

PR #1581, opened 2019-12-07
Fix a problem vetur show red underline errors in files whose names en…

gitercn:
…d with upper case like "Vue" or "VUE"

Currently, if the Vue file name extension is end with lower case letters "*.vue", everything is fine. But if it's end with upper case letters like "Vue" or "VUE", vetur will show red underline errors in these files. I add `.toLowerCase()` before all the `.endsWith` to fix this problem.


octref:
I think it's an anti-pattern to have `.Vue`, `.VUE` suffixes. All official projects / docs use `.vue` extension. I'd take a PR that shows error for `.Vue` or `.VUE` suffixes rather than relaxing the rules.

PR #1634, opened 2020-01-12
Vti

### LINKED ISSUE #1149: Offer VLS through CLI
Recently I changed some rules of my prettier config file, so I have to lint my whole project, and I could format my javascript file via prettier cli, but there is not such command for my vue file, so is it possible to support such command to format the whole project?

octref:
Fix #1634
Fix #1149


garyo:
Would love to see this get released in VLS!

octref:
https://github.com/vuejs/vetur/tree/master/vti

garyo:
Interesting -- that's a sort of command-line version of VLS? How would you expect that to be used?

I'm using VLS with Emacs (I'm using eglot). Recently noticed the switch from `vue-language-server` to `vls`, and I'll be very happy to take out the big config block from my `.emacs` file when this is out.

octref:
VTI doesn't replace VLS. It's just `vue-language-server` npm package will be renamed to `vls`.

Issue #1316, opened 2019-05-31
Language Server: Problem with incorrect paths in Windows


Narretz:
- [x] I have searched through existing issues
- [x] I have read through [docs](https://vuejs.github.io/vetur)
- [x] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

## Info

- Platform: Windows
- Language Server: 0.0.56

## Problem

I'm using the vls together with Language Server Protocol in Sublime Text 3.
However, the problem doesn't look like it depends on these programs.

What happens is that VLS in some cases passes the wrong path to typescript, which makes it fail when trying to find a source file. For example:

https://github.com/vuejs/vetur/blob/753c27bf5c1e2f3afd0c6f79f0f86f2cdd060bcf/server/src/modes/script/javascript.ts#L110-L114

This produces an uri like `c:\Users\path\src\components\App.vue` but Typescript internally has this file stored as `C:/Users/path/src/components/App.vue`

octref:
> in some cases passes
Do you have a specific case that I can repro?

Can you repro by modifying https://github.com/octref/veturpack?

Narretz:
I still don't have a  reproducible case, but I looked into this further I may have found the problem.

vls usually uses `getFsFilePath` to get the path to a file before it sends it to the relevant service, e.g. here: https://github.com/vuejs/vetur/blob/16e6f80bf7916d2db0ad2759d7f27a0afa625edd/server/src/modes/script/javascript.ts#L249

getFsFilePath runs vscode-uri.parse, which on Windows gives the following result:
`file:///C:/Users/Martin/dev/newnow/pdf_vue/src/components/HelloWorld.vue`
->
`c:\Users\path\src\components\App.vue`

As written above, this breaks typescript in my setup (Windows + Sublime Text + Sublime LSP) because previously, the tsserver has stored the sourceFiles like so: `c:/Users/path/src/components/App.vue`

Now there's another function called getFilePath:
https://github.com/vuejs/vetur/blob/16e6f80bf7916d2db0ad2759d7f27a0afa625edd/server/src/utils/paths.ts#L47

which returns exactly what we (I) need: 
`c:/Users/path/src/components/App.vue`

Any reason why `getFsFilePath` instead of `getFilePath` is used?

I noticed that in another function, getFilePath is used to see if a file exists in the service:
[...]

Issue #1589, opened 2019-12-12
Broken intellisense with typescript 3.7.3 (Vetur 0.23.0)


Niekvdm:
- [x] I have searched through existing issues
- [x] I have read through [docs](https://vuejs.github.io/vetur)
- [x] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

## Info

- Platform: Windown 10
- Vetur version: 0.23.0
- VS Code version: 1.40.2
- Extensions Enabled: Vetur only

## Problem

Build latest master source and named it for example 0.22.7.
This latest master version will be named 0.23.0 according to the changelog.

The issue is that within SFC .vue files (typescript) intellisense stops working.
![image](https://user-images.githubusercontent.com/3315401/70691872-7293b900-1cba-11ea-9d3a-94632479b5ab.png)

In the image above it showcases the issue, type definition on hover and intellisense (shift+space) are stuck on loading.

This also happened on Vetur 0.22.6 with useWorkspaceDependencies (TS 3.7.x) enabled.
Typescript 3.6.x works properly.

![image](https://user-images.githubusercontent.com/3315401/70692016-c56d7080-1cba-11ea-904d-07447a1efb80.png)

I've mentioned this issue before here:
https://github.com/vuejs/vetur/issues/1510#issuecomment-557410649

This is issue showcases the problem.

## Reproducible Case

Build from master branch including Typescript 3.7.x (3) and try to use intellisense or resolve types on hover.

octref:
Is it slow or is it completely broken?
Any error in Help -> Developer Tools or Output -> Vue Language Server?

Niekvdm:
> Is it slow or is it completely broken?
> Any error in Help -> Developer Tools or Output -> Vue Language Server?

Completely broken, first time you try to resolve something it breaks no errors.
F12 goto to definition breaks aswell and it will keep loading indefinitely.

Niekvdm:
@octref I've tried to captured it in a short video:
[Vetur_0.23.0_Typescript_3.7.3_broken_intellisense.MP4](https://drive.google.com/file/d/1srxNAC-5MmaWFrftWRHlFNGDxLHntvwO/view)

As shown in the video upon a fresh reload of vscode all is wel until you start typing and try to use intellisense everything breaks. unfortunately no errors are show in the console or any of the vscode logs.

pellelauritsen:
Is this the reason why v0.23 is not on VS Marketplace ?

[...]

Issue #1266, opened 2019-05-01
Setup Azure Devops CI


octref:
So I can run extension tests on Windows.

octref:
@yyx990803 Wondering if you can let me have hook access for this repo. I'm setting up Azure Devops since it's the only CI that can let me run extension tests on Windows.

yyx990803:
I think you already have admin access, does it not work?

octref:
OAuth failed for me but then I tried manually creating personal access tokens, and it worked. Thanks for taking a look!

octref:
Still a few tests to fix, but mostly looking good on Windows...

octref:
When running locally, here are the failing tests:

```
 1) Should find definition
       finds definition for this.msg:

      AssertionError [ERR_ASSERTION]: '/c:/Users/octref/Desktop/vetur/test/lsp/fixture/client/definition/Basic.vue' == '/C:/Users/octref/Desktop/vetur/test/lsp/fixture/client/definition/Basic.vue'
      + expected - actual

      -/c:/Users/octref/Desktop/vetur/test/lsp/fixture/client/definition/Basic.vue
      +/C:/Users/octref/Desktop/vetur/test/lsp/fixture/client/definition/Basic.vue

      at C:\Users\octref\Desktop\vetur\test\lsp\definition\basic.test.ts:45:10
      at Generator.next (<anonymous>)
      at fulfilled (C:\Users\octref\Desktop\vetur\dist\test\lsp\definition\basic.test.js:4:58)

  2) Should find definition
       finds definition for lodash:

      AssertionError [ERR_ASSERTION]: '/c:/Users/octref/Desktop/vetur/test/lsp/fixture/node_modules/@types/lodash/index.d.ts' == '/C:/Users/octref/Desktop/vetur/test/lsp/fixture/node_modules/@types/lodash/index.d.ts'
      + expected - actual

      -/c:/Users/octref/Desktop/vetur/test/lsp/fixture/node_modules/@types/lodash/index.d.ts
      +/C:/Users/octref/Desktop/vetur/test/lsp/fixture/node_modules/@types/lodash/index.d.ts

      at C:\Users\octref\Desktop\vetur\test\lsp\definition\basic.test.ts:45:10
      at Generator.next (<anonymous>)
      at fulfilled (C:\Users\octref\Desktop\vetur\dist\test\lsp\definition\basic.test.js:4:58)


  3) Should find definition
       finds definition for Vue#data:

[...]

PR #1662, opened 2020-01-21
Fix Windows paths.

mattn:
vscode-uri use lower-case drive-letter. So convert it always in getFilePath()

Fixes https://github.com/vuejs/vetur/issues/1316

### DIFF
--- a/server/src/services/vls.ts
+++ b/server/src/services/vls.ts
@@ -1,4 +1,5 @@
 import * as path from 'path';
+import { getFileFsPath } from '../utils/paths';
 
 import {
   DidChangeConfigurationParams,
@@ -198,7 +199,7 @@ export class VLS {
 
       changes.forEach(c => {
         if (c.type === FileChangeType.Changed) {
-          const fsPath = Uri.parse(c.uri).fsPath;
+          const fsPath = getFileFsPath(c.uri);
           jsMode.onDocumentChanged!(fsPath);
         }
       });
@@ -349,8 +350,8 @@ export class VLS {
         if (this.workspacePath && ref[0] === '/') {
           return Uri.file(path.resolve(this.workspacePath, ref)).toString();
         }
-        const docUri = Uri.parse(doc.uri);
-        return Uri.file(path.resolve(docUri.fsPath, '..', ref)).toString();
+        const fsPath = getFileFsPath(doc.uri);
+        return Uri.file(path.resolve(fsPath, '..', ref)).toString();
       }
     };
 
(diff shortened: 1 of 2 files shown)


octref:
I'm traveling and don't have a Windows machine at hand, but I was trying to fix Windows CI (https://github.com/vuejs/vetur/issues/1266#issuecomment-490295222), so I pushed two branches to see what would happen:

- https://github.com/vuejs/vetur/tree/windows-ci
- https://github.com/vuejs/vetur/tree/pine/1662-windows

mattn:
Any update?

server/src/services/vls.ts
```diff
@@ -200,7 +201,7 @@ export class VLS {
 
       changes.forEach(c => {
         if (c.type === FileChangeType.Changed) {
-          const fsPath = Uri.parse(c.uri).fsPath;
+          const fsPath = getFileFsPath(c.uri);
```
octref: Did you mean to change to `getFilePath`? How is this change different than `getFileFsPath`'s implementation?

server/src/utils/paths.ts
```diff
@@ -51,7 +51,9 @@ export function getFilePath(documentUri: string): string {
   const IS_WINDOWS = platform() === 'win32';
   if (IS_WINDOWS) {
     // Windows have a leading slash like /C:/Users/pine
-    return Uri.parse(documentUri).path.slice(1);
+    // vscode-uri use lower-case drive letter
+    // https://github.com/microsoft/vscode-uri/blob/95e03c06f87d38f25eda1ae3c343fe5b7eec3f52/src/index.ts#L1017
+    return Uri.parse(documentUri).path.replace(/^\/[a-zA-Z]/, (s: string) => s.slice(1).toLowerCase());
```
octref: This change would only make drive letters lowercase for you...Does that fix your issue?

Or is your issue a bug on TS 3.7.3? See https://github.com/vuejs/vetur/issues/1589#issuecomment-583991616 which includes some changes upstream that handles drive letters differently.
If upgrading to TS 3.7.5 (with `vetur.useWorkspaceDependencies`) fixes your issue, then this isn't the correct fix.

Narretz: I've tested this on Windows and it actually fixes my path issues I tested from master branch which includes ts 3.7.5 (message was vue-ls: Loaded bundled typescript@3.7.5). Without this line change, vetur didn't work.

server/src/services/vls.ts
mattn: It is unification of processing. In the future getFileFsPath will allow you to make unified fine-tuning(or fixing).

server/src/utils/paths.ts
octref: Thanks, rebasing on master and will release once CI goes through.

mattn:
Thank you

octref:
Thank you as well!

PR #1495, opened 2019-11-10
Add more typescript language service features and fix more bugs

yoyo930021:
1. Add suggestion diagnostics and can fix unused import.
2. Add Organize Imports code actions. Fix #843 
3. Diagnostics will show quick fix in VScode.
Screen Shot 2019-11-11 at 1 29 43 AM
Screen Shot 2019-11-11 at 1 44 50 AM

4. try to add #871  in typescript.
5. Add support resolveJsonModule

PR #1704, opened 2020-02-13
Use CodeAction over command

mjbvz:
Switch the code action provider to use CodeActions instead of commands. CodeActions work better with VS Code's `Refactor` commands. They also let us avoid an extra command invocation


rchl:
@mjbvz While the `vetur.applyWorkspaceEdits` was refactored to be a pure code action with edit, the `vetur.chooseTypeScriptRefactoring` is still implemented in the extension so doesn't work in other editors.

Can you say something about that? Is there a reason you couldn't refactor that also?
I understand that for performance reasons you might want to delay computations and thus use the command but I'm not quite  sure I understand why it can't be a pure server command instead.

yoyo930021:
> @mjbvz While the `vetur.applyWorkspaceEdits` was refactored to be a pure code action with edit, the `vetur.chooseTypeScriptRefactoring` is still implemented in the extension so doesn't work in other editors.
> 
> Can you say something about that? Is there a reason you couldn't refactor that also?
> 
> I understand that for performance reasons you might want to delay computations and thus use the command but I'm not quite sure I understand why it can't be a pure server command instead.

I will study this question when I refactor #1495. 
I will try to use a pure server command instead.

Issue #977, opened 2018-11-24
vue-language-server requires a config to be provided (probably shouldn't)


pearofducks:
- [ x ] I have searched through existing issues
- [ x ] I have read through [docs](https://vuejs.github.io/vetur)
- [ x ] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

## Info

- Platform: macOS 10.14.1
- Vetur version: `vue-language-server@0.0.35`
- VS Code version: `n/a`

## Problem

[This code in vue-language-server](https://github.com/vuejs/vetur/blob/master/server/src/modes/vue/index.ts#L15) assumes that `config.vetur.completion` exists.

Non-VSCode implementations (namely LSP for Sublime) may not initialize `config`, thus this check crashes and Vue completions are often not provided.

Is it possible for vue-language-server to initialize config with some sane defaults? Or to safely attempt to access this rather deep attribute on the config object?

## Reproducible Case

Don't initialize `config` when starting the language server, see lots of crashes because of `undefined`.

octref:
This makes sense. #745 is outdated. I need to revive it.

rwols:
Using the initializationOptions dict in the initialize request seems to be the most editor-agnostic way for passing server-specific options. I recommend going down that route if you want to transparently support other editors.

bstaletic:
Why am I getting errors regarding missing configuration (see #1462) even with the following configuration?

```
{
    "vetur.useWorkspaceDependencies": False,
    "vetur.completion.autoImport": True,
    "vetur.completion.scaffoldSnippetSources": {
      "workspace": "💼",
      "user": "🗒️",
      "vetur": "✌"
      },
    "vetur.completion.tagCasing": "kebab",
    "vetur.grammar.customBlocks": { "docs": "md", "i18n": "json" },
    "vetur.validation.template": True,
    "vetur.validation.style": True,
    "vetur.validation.script": True,
    "vetur.format.enable": True,
    "vetur.format.options.tabSize": 2,
    "vetur.format.options.useTabs": False,
    "vetur.format.defaultFormatter.html": "prettyhtml",
    "vetur.format.defaultFormatter.css": "prettier",
    "vetur.format.defaultFormatter.postcss": "prettier",
    "vetur.format.defaultFormatter.scss": "prettier",
    "vetur.format.defaultFormatter.less": "prettier",
    "vetur.format.defaultFormatter.stylus": "stylus-supremacy",
[...]

PR #1705, opened 2020-02-13
Fixes errors when no initialization options are given to vue-language-server

hacktivista:
This solves #977 and builds upon the work of #1634 

Without this fixes I wasn't able to initialize the server for Emacs Eglot without errors, it should solve the same issues on other clients.

Notice I'm using the "optional chaining operator" on the `VLSFullConfig` assignation (given that this project uses Babel).


octref:
👍  Thanks and sorry for late response.
Rebasing and making some small changes.

PR #1708, opened 2020-02-14
update prettyhtml other settings

bestvow:
add options wrapAttributes and sortAttributes


octref:
This just shows how you can use prettier options. You don't have to list all options. Still, thanks for your PR!

PR #1710, opened 2020-02-14
fix: enable markdown for library documentation

### LINKED ISSUE #1706: Tags and Attributes descriptions are not being parsed as markdown anymore.

## Info

- Platform: Win
- Vetur version: 0.23.0
- VS Code version: 1.42.0

## Problem

Tag and attributes auto-completions are not accepting markdown anymore.

This is possibly due to Vetur using [`MarkedString`](https://vshaxe.github.io/vscode-extern/vscode/MarkedString.html) from `vscode-languageserver-types`, in the following line:

https://github.com/vuejs/vetur/blob/73a0854b9eba3311951546a1e82a6832b523acbb/server/src/modes/template/services/htmlHover.ts#L3

This type is deprecated and [`MarkdownString`](https://vshaxe.github.io/vscode-extern/vscode/MarkdownString.html#MarkdownString) should be used instead.

## Reproducible Case

This is reproducible in any repository that has a framework with the tags and attributes provider.
Here's a screenshot of the issue that is happening: 
image

nekosaur:
show documentation strings from providers as markdown

closes #1706


nekosaur:
I tried to replicate the test failures on OS X with the same node version but they all passed. Any ideas @octref ?

octref:
Probably due to flakey CI. Now it should be better. Sorry for my late response.

octref:
Thanks for fixing this 👍

PR #1736, opened 2020-02-27
[WIP] feat(completion): improve auto complete in html template

pikax:
Improving auto completion on html, based on the component definition.

## Implemented
- [x] Provide deep member properties

## TODO 
- [ ] Support auto-completion when is expression

image

image
image

image

example:
```vue
<template>
  <div>
    <p v-if="obj.ourdata"></p>
  </div>
</template>

<script>
import { defineComponent } from "@vue/composition-api";

export default defineComponent({
  data() {
    return {
      mydata: "1",
      obj: {
        ourdata: 2,
        supa: {
          a: "a",
          n: 1122
        }
      }
    };
  },
  computed: {
    test() {
      return 1;
    },
    testObj() {
      return {
        a: 1,
        bb: {
          a: "2"
        }
      };
    }
  }
});
</script>
```

### DIFF
--- a/server/src/modes/script/componentInfo.ts
+++ b/server/src/modes/script/componentInfo.ts
@@ -5,8 +5,10 @@ import {
   ComputedInfo,
   DataInfo,
   MethodInfo,
-  ChildComponent
+  ChildComponent,
+  MemberInfo
 } from '../../services/vueInfoService';
+import { CompletionItemKind, combineClientFeatures } from 'vscode-languageserver';
 import { getChildComponents } from './childComponents';
 import { T_TypeScript } from '../../services/dependencyService';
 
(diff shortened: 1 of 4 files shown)


ktsn:
Hmm, I don't think we should manually analyze data properties...
We should utilize type info to do that instead so that the behavior would be the same as normal typescript language feature and we don't have to reinvent the wheel.
And I had made such implementation earlier on this PR. https://github.com/vuejs/vetur/pull/1446

pikax:
If there's a way to get the autocomplete correctly from typescript, I completely agree on using that.

Checked your PR(still need time to understand it), only get the first level auto completion, I reckon is the https://github.com/vuejs/vetur/blob/271dbfcd9b1be62f847e6edf0041ee0775ee2f91/server/src/modes/script/componentInfo.ts#L197-L202

Is not enough to get the the info for 
https://github.com/vuejs/vetur/blob/271dbfcd9b1be62f847e6edf0041ee0775ee2f91/server/src/modes/template/services/vueInterpolationCompletion.ts#L49-L55

To build the proper typing.


because as soon as you do:
```html
<p v-if="obj.ou">
```
The  templateService will return undefined.

https://github.com/vuejs/vetur/blob/271dbfcd9b1be62f847e6edf0041ee0775ee2f91/server/src/modes/template/interpolationMode.ts#L119

ktsn:
You are seeing the current implementation that is even left in the PR for backward compatibility.

The main part of the completion is this line.
https://github.com/vuejs/vetur/pull/1446/files#diff-62e73e93316645d3d052a2638e515905R119
Just calling TS auto completion logic so that we can complete anything if the target is properly typed.

The nested property completion is confirmed by a test case in the PR as well.
https://github.com/vuejs/vetur/pull/1446/files#diff-01bc940b65f71e51cdbf9185aecefa8aR4

pikax:
> The nested property completion is confirmed by a test case in the PR as well.
> https://github.com/vuejs/vetur/pull/1446/files#diff-01bc940b65f71e51cdbf9185aecefa8aR4

That example works, the example I was using, the typescript is not actually working -.-''

What's is needed for that PR to be merged?

Issue #1711, opened 2020-02-15
yarn@berry support.


Sombian:
- [x] I have searched through existing issues
- [x] I have read through [docs](https://vuejs.github.io/vetur)
- [x] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

## Info

- Platform: windows 10 x64 pro
- Vetur version: latest (0.23.0)
- VS Code version: latest

## Problem

![image](https://user-images.githubusercontent.com/23397550/74586547-fc019780-502b-11ea-8cb6-50e0319e07ce.png)

as you can see from the picture above, it warns user that module has not found (this also includes built-in) whilst its not.
never had this problem without yarn@berry.

## Reproducible Case

on existing or freshly forged project, with ``yarn@berry`` either globally or scoped install,
run ``yarn install``.
then in ``.vue`` files, try import any of modules (built-in or external, native ``(java|type)script`` works fine though)

Qwertronix:
A way to have the VLS use the tsserver and tsc pointed at by the `"typescript.tsdk"` setting in vscode could provide at least a temporary solution to this until more indepth support for PnP can be implemented. 

As a workaround in my current project I'm adding `nodeLinker: node-modules` to `.yarnrc.yml`, which disables PnP entirely and removes a lot of the benefits of `yarn@berry`.

Sombian:
> A way to have the VLS use the tsserver and tsc pointed at by the `"typescript.tsdk"` setting in vscode could provide at least a temporary solution to this until more indepth support for PnP can be implemented.
> 
> As a workaround in my current project I'm adding `nodeLinker: node-modules` to `.yarnrc.yml`, which disables PnP entirely and removes a lot of the benefits of `yarn@berry`.

then theres no point of using yarn@berry if i do so.
thanks for your suggest but id rather keep error as it wont cause error on script but workspace.

octref:
I'm not sure what I can do to fix this. What does yarn@berry do different than yarn v1 to modules? Does TS support it?

Seally:
Yarn v2 defaults to PnP (which is in v1 too, but not enabled by default), which basically means `node_modules` is gone (unless the `node-modules` linker is used) and module resolutions are now handled by the `.pnp.js` file generated at the project root. The installed modules themselves are generally stored in zip archives, which reside in `<root>/.yarn/cache` by default for Yarn v2. The `.pnp.js` file I believe also comes with the virtual file system needed to access stuff within the archives.

[...]

PR #1737, opened 2020-02-28
fix(#1711) Get workspace typescript with the tsdk path (yarn@berry compat)

AlexandreBonaventure:
Naive attempt (proof of concept) to fetch TS runtime based on the `vscode.typescript.tsdk` setting with a fallback to `node_modules`. This will allow to use yarn@berry patched TS runtime as stated in #1711
I'd like to get some feedback/help to finalize this PR

### DIFF
--- a/server/src/services/dependencyService.ts
+++ b/server/src/services/dependencyService.ts
@@ -54,7 +54,7 @@ export class DependencyService {
 
   constructor() {}
 
-  async init(workspacePath: string, useWorkspaceDependencies: boolean) {
+  async init(workspacePath: string, useWorkspaceDependencies: boolean, tsSDKPath: string) {
     if (!useWorkspaceDependencies) {
       const tsModule = await import('typescript');
       console.log(`Loaded bundled typescript@${tsModule.version}.`);
@@ -67,7 +67,7 @@ export class DependencyService {
         module: tsModule
       };
     } else {
-      const workspaceTSPath = path.resolve(workspacePath, 'node_modules/typescript');
+      const workspaceTSPath = path.resolve(workspacePath, tsSDKPath || 'node_modules/typescript');
       let tsModule: T_TypeScript;
       let bundled = false;
       try {
--- a/server/src/services/dependencyService.ts
+++ b/server/src/services/dependencyService.ts
@@ -67,7 +67,8 @@ export class DependencyService {
         module: tsModule
       };
     } else {
-      const workspaceTSPath = path.resolve(workspacePath, tsSDKPath || 'node_modules/typescript');
+      const tsPath = tsSDKPath && path.join(tsSDKPath, '..');
+      const workspaceTSPath = path.resolve(workspacePath, tsPath || 'node_modules/typescript');
       let tsModule: T_TypeScript;
       let bundled = false;
       try {
--- a/server/src/services/dependencyService.ts
+++ b/server/src/services/dependencyService.ts
@@ -54,7 +54,7 @@ export class DependencyService {
 
   constructor() {}
 
-  async init(workspacePath: string, useWorkspaceDependencies: boolean, tsSDKPath: string) {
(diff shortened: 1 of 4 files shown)


server/src/services/vls.ts
```diff
@@ -97,7 +97,7 @@ export class VLS {
     this.workspacePath = workspacePath;
 
     await this.vueInfoService.init(this.languageModes);
-    await this.dependencyService.init(workspacePath, config.vetur.useWorkspaceDependencies);
+    await this.dependencyService.init(workspacePath, config.vetur.useWorkspaceDependencies, config.typescript.tsdk);
```
AlexandreBonaventure: I have no idea how to cleanly get the vscode `typescript.tsdk` setting. So this is really a guess, not tested obviously.

server/src/services/dependencyService.ts
```diff
@@ -67,7 +67,7 @@ export class DependencyService {
         module: tsModule
       };
     } else {
-      const workspaceTSPath = path.resolve(workspacePath, 'node_modules/typescript');
+      const workspaceTSPath = path.resolve(workspacePath, tsSDKPath || 'node_modules/typescript');
```
AlexandreBonaventure: Not sure if this kind of fallback actually works. We would have to know what is the default value of the `typescript.tsdk` setting

sbresin: The `typescript.tsdk` setting points directly to the `lib` folder, while we need to point to the parent to do `import(workspaceTSPath)`. This change makes it work.
If the `typescript.tsdk` is unset it defaults to `null`. So this would fallback to the old behaviour and try to load typescript from `node_modules/typescript`.

```suggestion
      const tsPath = tsSDKPath && path.join(tsSDKPath, '..');
      const workspaceTSPath = path.resolve(workspacePath, tsPath || 'node_modules/typescript');
```

sbresin:
hello. i have tested your changes and it's not working as it is. i added a comment how to make it work.

Additionaly, this doesn't work alone. While `yarn pnpify --sdk` sets up `.yarn/pnpify/typescript/lib/tsc.js` and `.yarn/pnpify/typescript/lib/tsserver.js` for us. In this case we also need `.yarn/pnpify/typescript/lib/typescript.js` with this content: 
```
#!/usr/bin/env node

const {createRequire, createRequireFromPath} = require(`module`);
const {resolve} = require(`path`);

const relPnpApiPath = "../../../../.pnp.js";

const absPnpApiPath = resolve(__dirname, relPnpApiPath);
const absRequire = (createRequire || createRequireFromPath)(absPnpApiPath);

// Setup the environment to be able to require typescript/lib/typescript.js
require(absPnpApiPath).setup();

// Defer to the real typescript/lib/typescript.js your application uses
module.exports = absRequire(`typescript/lib/typescript.js`);

```

this could easily be added to `@yarnpkg/pnpify` with a one line change.

AlexandreBonaventure:
@sbresin sounds good! thank you

merceyz:
I haven't tested this but this line might be a potential issue https://github.com/vuejs/vetur/blob/29122dd4a796d93cc0c9a52c0d7bd9ff039a5a83/server/src/modes/template/tagProviders/index.ts#L120

yoyo930021:
Fixed #1996

server/src/services/vls.ts
octref: The only issue is this doesn't handle such global settings:

```json
{
  "typescript.tsdk": "/usr/local/lib/node_modules/typescript/lib"
}
```

As seen in https://code.visualstudio.com/docs/typescript/typescript-compiling#_using-newer-typescript-versions

octref (changes requested):

octref (approved):

octref:
Now all good. I tested both workspace / global tsdk setting. @AlexandreBonaventure thanks for your work and sorry for late response.

Issue #1712, opened 2020-02-15
Jump to definition for Vue tags in template region


HuYong66:
In vue project,  we can't  jump to the  child components by click the child components tag.  However,  it's enabled in react project.

andrewisaburden:
@octref @yoyo930021 this can be closed, it's working.

PR #1760, opened 2020-03-07
feat: find correct child components' definition inside template region

hikerpig:
This PR will add support for #1712.

I notice that there is already `vueFileInfo.componentInfo.childComponents` provided by the implementation of Vue interpolation. But during definition-finding for html, a little case issue causes this functionality to fail, typically when component filenames adopt a PascalCase convention.


triforcely:
Are there any reasons why this PR was not merged?

PR #1788, opened 2020-03-16
Fix template interpolation in vfor when readonly array

yoyo930021:
When array is readonly, I will get `any` in `v-for` element.

### DIFF
--- a/server/src/services/typescriptService/bridge.ts
+++ b/server/src/services/typescriptService/bridge.ts
@@ -28,7 +28,7 @@ export declare const ${componentHelperName}: {
   ): any;
 };
 export declare const ${iterationHelperName}: {
-  <T>(list: T[], fn: (value: T, index: number) => any): any;
+  <T>(list: readonly T[], fn: (value: T, index: number) => any): any;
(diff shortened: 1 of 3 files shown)


server/src/services/typescriptService/bridge.ts
```diff
@@ -29,6 +29,7 @@ export declare const ${componentHelperName}: {
 };
 export declare const ${iterationHelperName}: {
   <T>(list: T[], fn: (value: T, index: number) => any): any;
+  <T>(list: readonly T[], fn: (value: T, index: number) => any): any;
```
ktsn: I believe just changing existing `T[]` overload is enough rather than adding a new one.

yoyo930021: I change `T[]` to `readonly T[] | T[]`.

ktsn: I mean just changing it to `readonly T[]` because it's super type of `T[]`

yoyo930021: Wow! I'm done.

Issue #1699, opened 2020-02-09
VTI crashes with exit code 1 without any logs


ktsn:
- [x] I have searched through existing issues
- [x] I have read through [docs](https://vuejs.github.io/vetur)
- [x] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

## Info

- Platform:  macOS 10.15.3
- VTI version: 0.0.2
- VS Code version: -

## Problem

`vti diagnostics` crashes without any error logs on a larger project.
- number  of `.vue` -> 134
- number of `.ts` -> 120

<img width="617" alt="Screenshot 2020-02-09 at 4 06 48 PM" src="https://user-images.githubusercontent.com/2194624/74098810-344e3500-4b57-11ea-9a95-62058083e2b0.png">

I couldn't catch any errors even though I set `uncaughtexception` and `unhandledrejection` even in vti source.

## Reproducible Case

It's pretty hard to reproduce as it looks depends on the size of a project (maybe memory allocation related?) and the reproducible project is confidential.
I've got some memory report in the attached zip file.

https://drive.google.com/file/d/1zUKepNdxCH7Imk2zR7bHvArcGQ8MDmwn/view?usp=sharing

- `Timeline.heaptimeline` Memory allocation timeline through `vti diagnostics`
- `Snapshot1.heapsnapshot` Snapshot just after starting `vti diagnostics`
- `Snapshot2.heapsnapshot` Snapshot after 2-3 seconds of starting `vti diagnostics`
- `Snapshot3.heapsnapshot` Snapshot just before `vti diagnostics` crashes

jackkoppa:
Hi @ktsn - I'm having the same exact experience. I adding logging & breakpoints to both `vti > cli.js`, and `vue-language-server > vls.js`, but there are no JS exceptions being thrown before it exits with `1`. In my case, for a small project, it will usually get through the 8 files that it needs to check, but fail right before the validation result is returned to `vti > cli.js` for the final file. Though there's some variation on when exactly the failure occurs, if I try the command 10 different times. And, every once in a while, it gets all the way through without errors

Will create a repro case repo

jackkoppa:
[...]

PR #1805, opened 2020-03-24
Pass correct PID to language server

TheBeastOfCaerbannog:
[VS Code Language Server sends `SIG_0` to parent process](https://github.com/microsoft/vscode-languageserver-node/blob/master/server/src/main.ts#L1899-L1911), to make sure it's still alive. However, calling `kill` with `SIG_0` also performs permission check, and since VTI passes 1 as processId - `kill` fails and language server thinks the parent process doesn't exist.
This PR suggests a fix to this problem by passing a real PID.

Resolves: #1699

### DIFF
--- a/vti/src/cli.ts
+++ b/vti/src/cli.ts
@@ -64,6 +64,7 @@ async function prepareClientConnection(workspaceUri: Uri) {
   const init: InitializeParams = {
     rootPath: workspaceUri.fsPath,
     rootUri: workspaceUri.toString(),
+    processId: process.pid,
     ...params
   } as InitializeParams;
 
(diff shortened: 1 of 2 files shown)


TheBeastOfCaerbannog:
Is it possible to merge this PR please? This bug makes VTI pretty much unusable :(

dimitor115:
@octref  @ktsn what about merging it? 😸 We are waiting for this fix!

jackkoppa:
Awesome, thank you @TheBeastOfCaerbannog! 🙏  Great debugging on this

octref:
Sorry I have been on leave.
Great work and thanks for your contribution 👍

PR #1822, opened 2020-04-03
feat: treat nuxt-buefy as Buefy framework

nogic1008:
[nuxt-buefy](https://github.com/buefy/nuxt-buefy) is the nuxt plugin developed by [Buefy](https://github.com/buefy) official team.
And, this plugin contains latest buefy package.

So we can think of it as one of the buefy frameworks, just like the relationship between `vuetify` and `@nuxtjs/vuetify`.


nogic1008:
Close this PR because #1866 contains this change

Issue #1212, opened 2019-04-19
Narrowed type by v-if should not be widened in v-on etc.


ktsn:
The current v-if narrowing implementation (#1208), there are some cases the narrowed type is widened to the original type.

```vue
<template>
  <div v-if="post">
    <!-- post is of type "Post" here -->

    <!-- post will be widened to "Post | null" in @click directive -->
    <button @click="onClick(post)">Click</button>
  </div>
</template>

<script lang="ts">
import Vue from 'vue'

interface Post {
  id: string
}

export default Vue.extend({
  data() {
    return {
      post: null as Post | null
    }
  },

  methods: {
    onClick(post: Post) {
      // ...
    }
  }
})
</script>
```

This is because v-on emits callback code and TypeScript widens the narrowed type in a callback. We may need to assign some local variable to retain narrowed types.

kimamula:
Is it possible to disable strict null checks of vti (or Vetur) only for v-on?

rchl:
As far as I can tell, this is the correct behavior. The click handler is evaluated on click, not at the time of rendering the template. So the data variable `post` can be null at the time of click handler being executed.

yoyo930021:
https://github.com/vuejs/vetur/issues/2504#issuecomment-735761621

SuspiciousLookingOwl:
Is this the same issue?

```vue
<template>
  <div>
    <div v-if="baz">
      <div v-for="x in [1, 2, 3, 4, 5]" :key="x">
        <!-- baz should be narrowed to `string`, but it's not -->
        {{ bar(baz) }}
      </div>
    </div>

    <div v-for="x in [1, 2, 3, 4, 5]" :key="x">
      <div v-if="baz">
        <!-- baz correctly narrowed to string if v-if is inside the loop -->
        {{ bar(baz) }}
      </div>
    </div>
  </div>
</template>

<script lang="ts">
import { Component, Vue } from "vue-property-decorator";

@Component
export default class AddToDialog extends Vue {
  get baz(): string | null {
    return "";
  }

  bar(x: string) {
    return x;
  }
}
</script>

```
![image](https://user-images.githubusercontent.com/32597776/100869510-bab17a00-34cf-11eb-8a88-459e1f8aa0b4.png)

dospunk:
I am having the same issue that @SuspiciousLookingOwl is having, and it only seems to happen with `"strict": true` in the tsconfig. Without strict being on the narrowing performs as expected.

[...]

PR #1824, opened 2020-04-05
 Apply template interpolation validation only to type-checked components

jackkoppa:
Addresses #1427 

Even when `vetur.experimental.templateInterpolationService: true`,
the nature of Vue 2 `this` reliance means that completely valid JS components
will still show errors in the template section,
because plain TS inference will not work without `Vue.extend({})` and certain type annotations
For this reason, Vetur doesn't attempt to provide type checking
in the <script> sections of components, unless one of the following is true:
1. component uses `lang="ts"`
2. components includes `// @ts-check` comment (& uses JSDoc for type annotations)
3. component imports a `.ts` external script using `src=""` (thus the type checking happens in that file)

Template interpolation should now work the same way (should only be applied in the above 3 scenarios),
to allow projects **with a mixture of type-checked & non-type-checked components**,
to still turn `templateInterpolationService` on, and get benefits in the type-checked components,
without getting lots of "incorrect" errors in non-type-checked components


server/src/modes/template/interpolationMode.ts
```diff
@@ -353,6 +355,36 @@ export class VueInterpolationMode implements LanguageMode {
...
+
+    if (!languageServiceIncludesFile(service, scriptUri) || !program) {
+      return false;
+    }
+
+    const fileFsPath = getFileFsPath(scriptUri);
+    const sourceFile = program.getSourceFile(fileFsPath);
+
+    // TODO: Raise issue in TS repo, to mark `checkJsDirective` as externally available
+    // per discussion on AST viewer: https://github.com/dsherret/ts-ast-viewer/issues/65#issuecomment-605706743
```
jackkoppa, line 379: In looking at the AST viewer, for both `checkJsDirective` and `scriptKind`, discovered that both properties are marked as `/* @internal */`; we can see that in the TS source code [here](https://github.com/microsoft/TypeScript/blob/master/src/compiler/types.ts#L3010), which is why `as any` is required.

However, as @dsherret points out [here](https://github.com/dsherret/ts-ast-viewer/issues/65#issuecomment-605706743), it does appear that [`CheckJsDirective`](https://github.com/microsoft/TypeScript/blob/master/lib/typescript.d.ts#L1568-L1570) is intended to be available in the public API ([`ScriptKind`](https://github.com/microsoft/TypeScript/blob/master/lib/typescript.d.ts#L2786-L2799) seems the same). And importantly, both are certainly available on `sourceFile` during runtime. I'll be opening an issue in the TypeScript repo, and hopefully they can either be exposed in types, or we can learn an alternative way to access both pieces of information

jackkoppa:
☝️ noted that the build failed on Windows; will run locally on a Windows machine to fix the file path resolution

jackkoppa:
Checks now pass, and would love to discuss with @ktsn or @octref, whenever someone is available - I'm sure I missed things about how these checks could be done more easily. Thanks!

(I'm hopeful this could get us to the point where we could consider moving `templateInterpolationService` out from the `experimental` flag, and possibly turning on by default, to start exposing more devs to the feature. Given how well it's worked for us in our type-checked files - we have yet to run into any bugs/incorrect errors in those. Though I know that's a later discussion)

cc @F0rsaken

server/src/modes/template/interpolationMode.ts
jackkoppa: Got an update from @DanielRosenwasser (much obliged 🙇), that it's intentional that both properties aren't exposed in the `.d.ts` files: https://twitter.com/drosenwasser/status/1247047515951652864

So what I'm probably missing are exposed API methods to access the same info, meaning I have to dig a bit deeper into everything currently available on the `typescript` module itself. I haven't been able to find something suitable yet; if I still can't after some more research, then I'd file a TS issue to expose APIs for these

We probably shouldn't merge this until we know whether APIs already exist or will need to be created, because as Daniel points out, the type cast is probably only good for prototype-level projects. But everything else can be reviewed just fine, if we're OK assuming "and somehow, here we get whether `@ts-check` is present, and somehow here, we know the scriptKind"

jackkoppa: This is the closest I've found to a single utility method that gives us what we want, using `SourceFile`: [`isUncheckedFile`](https://github.com/microsoft/TypeScript/blob/master/src/services/completions.ts#L306-L308)

F0rsaken:
> Checks now pass, and would love to discuss with @ktsn or @octref, whenever someone is available - I'm sure I missed things about how these checks could be done more easily. Thanks!
> 
> (I'm hopeful this could get us to the point where we could consider moving `templateInterpolationService` out from the `experimental` flag, and possibly turning on by default, to start exposing more devs to the feature. Given how well it's worked for us in our type-checked files - we have yet to run into any bugs/incorrect errors in those. Though I know that's a later discussion)
> 
> cc @F0rsaken

I seriously can't wait for it to go live and I can start using `interpolationService` in mixed projects! Thanks fo contribution @jackkoppa!

ktsn:
Thank you for making this working @jackkoppa. Please let me give some time to take a look at this as I'm currently kind of overwhelmed by my day job 😅 

> I'm hopeful this could get us to the point where we could consider moving templateInterpolationService out from the experimental flag, and possibly turning on by default, to start exposing more devs to the feature. 

This PR definitely makes it much closer to the point removing experimental flag but I would say we also need to solve #1212, which would be pretty annoying for users, before do that.

jackkoppa:
Thanks @ktsn! Totally understand that you've got a lot going on. 

> but I would say we also need to solve #1212, which would be pretty annoying for users, before do that 

In looking at #1208, I'm not confident I'd be able to understand/help much with #1212, but it's good to know that's another high priority for template interpolation. I'll try to take a look if I get a chance at some point

jackkoppa:
Hi @octref - great to see [another Vetur release](https://github.com/vuejs/vetur/releases/tag/v0.26.0) happen!

I can get back to this, to resolve conflicts and, optionally, to open the TS repo issue I mentioned to expose an API for [`isUncheckedFile`](https://github.com/microsoft/TypeScript/blob/master/src/services/completions.ts#L311). Anything else you'd like to see before this is merge-able?

rchl:
@jackkoppa I'm working on projects that are using SFC written in JS where I don't use `Vue.extend` but I have global shim to set the *.vue type correct:
```ts
declare module '*.vue' {
    import Vue from 'vue';
    export default Vue;
}
```

And I would like this to continue working without having to add `@ts-check` in every Vue file.

octref:
> Anything else you'd like to see before this is merge-able?

Sorry I haven't had a close look yet 😅 Will continue tomorrow. Planning to finish all other PRs before doing multiroot and sub-folder support. Just resolving the conflict is fine for now 👍

jackkoppa:
Thanks @octref, will work on resolving the conflicts

@rchl, interesting - I'm really surprised that that works effectively with no JSDoc annotations & no TS files, given things like computed props. Any sample repo you might be able to link to? The good news is that the TS method we'd hope to use if made available, `isUncheckedFile`, also checks whether [`compilerOptions.checkJs`](https://github.com/microsoft/TypeScript/blob/c9f05af2c870e785bb172286b8a946caf219f5d6/src/compiler/utilities.ts#L6520) is enabled, in which case not every file would need `@ts-check`; just defined once in tsconfig

Likely won't work as this PR stands, so would love to see a sample repo?

rchl:
> @rchl, interesting - I'm really surprised that that works effectively with no JSDoc annotations

I did not say that it works without JSDoc annotations. :)
I have `checkJs` enabled in `tsconfig.json`, I have the aforementioned shim and I'm adding JSDoc annotations everywhere where it's needed.

I don't have a public repo to show but here is one of the (redacted) components:
```vue
<template>
    <div>
        <p>{{ headerText }}</p>
        <p v-if="enabledFields.includes('name')">Something</p>
    </div>
</template>

<script>
import fieldName from '~/components/field-name.vue';

export default {
    components: {
        fieldName,
    },
    props: {
        /** @type {import('vue').PropOptions<string[]>} */
        enabledFields: {
            type: Array,
            required: true,
            validator: values => values.every(v => ['a', 'b', 'c'].includes(v)),
        },
        headerText: {
            type: String,
            default: '',
        },
    },
    computed: {
        /** @return {string} */
        something() {
            return 'sss'
        },
    },
    methods: {
        /**
         * @param {MouseEvent} event
         * @param {TIDAL.MenuItem} item
         */
        handleClick(event, item) {
            event.preventDefault();
        },
    },
};
</script>
```

The `tsconfig.json` looks more or less like this:
```js
{
    "compilerOptions": {
        "baseUrl": ".",
        "target": "esnext",
        "module": "esnext",
        "esModuleInterop": true,
        "moduleResolution": "node",
        "noEmit": true,
        "allowJs": true,
        "checkJs": true,
        "allowSyntheticDefaultImports": true,
        "strict": true,
        "resolveJsonModule": true,
        "lib": [
            "dom",
            "esnext",
        ],
    }
}
```

octref:
Can you tick "allow edits from maintainers", or do a rebase off origin/master? Thanks 👍

server/src/modes/template/interpolationMode.ts
yoyo930021: Hi @jackkoppa 
I study it.
I think we can convert to `isCheckJsEnabledForFile` and `isSourceFileJS` from `isUncheckedFile`.
https://github.com/microsoft/TypeScript/blob/06a2210eb53cfce968dc855fcc13738d95d4556a/src/services/completions.ts#L324
Although they too are not directly open in the definition.
But I think it's better than relying on this attribute.
We just need to make sure the program doesn't crash when these two methods are removed.

server/src/utils/paths.ts
```diff
@@ -62,3 +63,8 @@ export function getFilePath(documentUri: string): string {
 export function normalizeFileNameToFsPath(fileName: string) {
   return Uri.file(fileName).fsPath;
 }
+
+export function resolveUriAndPaths(documentUri: string, ...pathSegments: string[]) {
+  const resolvedPath = path.resolve(getFilePath(documentUri), ...pathSegments);
+  return Uri.file(resolvedPath).toString();
```
yoyo930021, line 69: Replace `.toString()` to `.fsPath`?

server/src/services/vueInfoService.ts
```diff
@@ -82,4 +82,8 @@ export class VueInfoService {
     });
     return this.vueFileInfo.get(getFileFsPath(doc.uri));
   }
+
+  getImportedScripts(doc: TextDocument): string[] {
```
yoyo930021, line 86: This function isn't here.
We only need to pass `documentRegions ` to `VueInterpolationMode`.
Just like `HTMLMode`.

yoyo930021 (commented):
LGTM, this PR can let us enable `vetur.experimental.templateInterpolationService` on defauly.

HaimCandiTech:
@jackkoppa 
Any updates on when will it be finally merged?

jackkoppa:
Hi all - it's been a couple years since I've worked closely with Vue + TypeScript in my day-to-day work, and so it'll be a little tricky for me to pick this back up / resolve conflicts / make the improvements mentioned in the comments. When I had left off with this a couple years ago, the PR did work, but I also imagine with continued Vue + Vetur changes, there might be tweaks needed. 

I don't want to contribute to long-stale PRs, so I'll be closing this for the time being. If I do end up in a Vue + TS codebase in the near future & still experience a similar issue, I'd come back to this, and anyone else is welcome to pick up the branch as well. Thanks all!

PR #1839, opened 2020-04-12
improvement: improve performance

IWANABETHATGUY:
i found a big performance issue when open a big vue project, and fix it , pass all test.

### DIFF
--- a/server/src/services/typescriptService/sourceMap.ts
+++ b/server/src/services/typescriptService/sourceMap.ts
@@ -274,7 +274,7 @@ function updateOffsetMapping(node: TemplateSourceMapNode, isThisInjected: boolea
   }
   const toFiltered = to as number[];
   if (isThisInjected) {
-    toFiltered.splice(nodeToStart, 5);
+    toFiltered.splice(nodeToStart, 'this.'.length);
   }
   const mapping = fillIntermediate
     ? from.map((from, i) => [from, toFiltered[i]])
--- a/server/src/services/typescriptService/sourceMap.ts
+++ b/server/src/services/typescriptService/sourceMap.ts
@@ -258,23 +258,19 @@ export function mapBackRange(fromDocumnet: TextDocument, to: ts.TextSpan, source
 function updateOffsetMapping(node: TemplateSourceMapNode, isThisInjected: boolean, fillIntermediate: boolean) {
   const nodeFromStart = node.from.start;
   const nodeToStart = node.to.start;
-  const from = [...Array(node.from.end - node.from.start + 1).keys()];
-  const to: (number | undefined)[] = [...Array(node.to.end - node.to.start + 1).keys()];
-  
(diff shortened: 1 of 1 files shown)


IWANABETHATGUY:
the same project, after open a few .vue file, i profile it.
**current version:**
Xnip2020-04-12_15-24-31
**this_pr version**
Xnip2020-04-12_14-59-48
**current updateOffsetMapping**
 
Xnip2020-04-12_15-25-14

**pr updateOffsetMapping**
Xnip2020-04-12_14-59-34

IWANABETHATGUY:
i hope this pr can be merged, so more people may benefit this improve @octref

ktsn:
If the perf bottle neck is `toFiltered`, can we only change it? Looks like inlining `for` loop for `to` and `this.`.length are unnecessary?

IWANABETHATGUY:
> If the perf bottle neck is `toFiltered`, can we only change it? Looks like inlining `for` loop for `to` and `'this.'.length` are unnecessary?

Because the string literal length `this.` can be calculated before runtime. This will not cause extra overhead . I think that if there is no particular impact on readability, performance should be the most important consideration.

ktsn:
According to your screen shot, there are no particular differences between them. 

- for loop
  - 6.9ms vs. 7.6ms
- `'this.'.length`
  - 1.4ms vs. 1.3ms


So I think we should keep them. We intentionally wrote as `'this.'.length` to indicate what the offset is for.

IWANABETHATGUY:
ok, i will do that

server/src/services/typescriptService/sourceMap.ts
```diff
@@ -256,24 +256,26 @@ export function mapBackRange(fromDocumnet: TextDocument, to: ts.TextSpan, source
...
      * The case such as `foo` mapped to `this.foo`
      * Both `|this.foo` and `this.|foo` should map to `|foo`
      * Without this back mapping, mapping error from `this.bar` in `f(this.bar)` would fail
      */
-    node.offsetBackMapping[node.to.start] = node.from.start + 'this.'.length;
+    node.offsetBackMapping[nodeToStart] = nodeFromStart + 'this.'.length;
+  }
+  const toFiltered = to as number[];
+  if (isThisInjected) {
+    toFiltered.splice(nodeToStart, 5);
```
ktsn, line 277:
```suggestion
    toFiltered.splice(nodeToStart, 'this.'.length);
```

For consistency.

IWANABETHATGUY:
@ktsn complete that

IWANABETHATGUY:
@ktsn  please don't publish now, i found this pr could cause `.vue` file outline issue, i will check that out

IWANABETHATGUY:
@ktsn i'm sorry, it's not a bug, it's my vscode config issue,  it's ok now .👍

PR #1841, opened 2020-04-12
fix: VTI - Wrong first arg

rafalt:
I fix a situation when you write incorrect first args.
Original version no display message.
I change the condition to display message always when not run any other action.


octref:
Cool, thank you 👍

PR #1866, opened 2020-04-22
feat: remove buefy-helper-json and use runtime pkg info

jtommy:
also added support through nuxt-buefy package

### DIFF
--- a/server/src/modes/template/tagProviders/index.ts
+++ b/server/src/modes/template/tagProviders/index.ts
@@ -6,7 +6,6 @@ import {
   elementTagProvider,
   onsenTagProvider,
   bootstrapTagProvider,
-  buefyTagProvider,
   gridsomeTagProvider,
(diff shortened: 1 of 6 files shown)


server/src/modes/template/tagProviders/index.ts
```diff
@@ -73,9 +71,12 @@ export function getTagProviderSettings(workspacePath: string | null | undefined)
     if (dependencies['bootstrap-vue']) {
       settings['bootstrap'] = true;
     }
-    if (dependencies['buefy']) {
+    if (dependencies['buefy'] || devDependencies['vuetify']) {
```
ygorre, line 74: Possible typo. Should be "buefy" instead of "vuetify".

jtommy:
@ygorre thanks

octref:
LGTM, sorry for the delay

PR #1867, opened 2020-04-23
update description about prettier

ylc395:
intent for Vue tags has been supported. see https://prettier.io/docs/en/options.html#vue-files-script-and-style-tags-indentation


octref:
Thank you 👍

PR #1897, opened 2020-05-07
Fix #1341 - Make sure initializationOptions is at least empty object

### LINKED ISSUE #1341: initializationOptions: Cannot read property 'config' of undefined
Hi, i'm trying to use vue-language-server with vim and LanguageClient-neovim and it does not work.

I traced bug to this error:
```
00:54:01 WARN unnamed src/language_server_protocol.rs:2132 Failed to start language server automatically. Error: Failure { jsonrpc: Some(V2), error: Error { code: InternalError, message: "Request initialize failed with message: Cannot read property \'config\' of undefined", data: None }, id: Num(1) }
```
https://github.com/autozimu/LanguageClient-neovim/blob/next/src/language_server_protocol.rs#L897

Looks like initializationOptions can be null, but vue-language-server does not take this into account here:

https://github.com/vuejs/vetur/blob/82d16031f58e0f54fde1fce94cf2c9518fe5080e/server/src/services/vls.ts#L83

It looks like trivial issue, can you fix it or should i make pull-request?

zulus:
Simple change that allows me to register vls in eclipse LSP4E


server/src/config.ts
```diff
@@ -74,7 +74,7 @@ export interface VLSFullConfig extends VLSConfig {
   stylusSupremacy?: any;
 }
 
-export function getDefaultVLSConfig(): VLSConfig {
+export function getDefaultVLSConfig(): VLSFullConfig {
```
zulus, line 77: Eclipse LSP4E wasn't able start because normally this field just not exists.

server/src/services/vls.ts
```diff
@@ -83,6 +83,7 @@ export class VLS {
   }
 
   async init(params: InitializeParams) {
+    params.initializationOptions = params.initializationOptions || {};
```
zulus, line 86: Required to use VLS in Eclipse LSP4E, probably other simple implementation also. This field normally not exists.

server/src/config.ts
```diff
@@ -118,6 +118,18 @@ export function getDefaultVLSConfig(): VLSConfig {
       experimental: {
         templateInterpolationService: false
       }
-    }
+    },
+    css: {},
```
zulus, line 122: VLS just forward config.emmet and others to sub-services. Even undefined, after this sass `<style lang="scss">` is operational.

octref:
No longer applies, and should already be fixed by #1705.

server/src/config.ts
octref: OK, will take this PR then

server/src/services/vls.ts
octref: No longer needed since we now use `?.` to access it

octref:
Thank you and sorry for the delay 👍

PR #1921, opened 2020-05-18
feat: detect tags from @nuxt/components

pi0:
Hi :) 

This PR adds integration with [nuxt/components](https://github.com/nuxt/components) (nuxt/components#14). I also did some refactors for `nuxtTags.ts` and also checking `devDependencies`. `nuxt-legacy` is no longer valid and `nuxt-ts` is installed alongside with nuxt dependency.

image

Locally working. The only issue is that I don't know if we can somehow reload vetur tags or not?

### DIFF
--- a/server/src/modes/template/tagProviders/index.ts
+++ b/server/src/modes/template/tagProviders/index.ts
@@ -96,13 +96,7 @@ export function getTagProviderSettings(workspacePath: string | null | undefined)
       // and enable Quasar later below in the for()
       dependencies['quasar-framework'] = '^0.0.17';
     }
-    if (
-      dependencies['nuxt'] ||
-      dependencies['nuxt-legacy'] ||
-      dependencies['nuxt-edge'] ||
-      dependencies['nuxt-ts'] ||
-      dependencies['nuxt-ts-edge']
-    ) {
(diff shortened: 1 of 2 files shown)


octref:
Thanks and sorry for the delay.

> The only issue is that I don't know if we can somehow reload vetur tags or not?

What do you mean by that? Updating tag definitions at run-time?

pi0:
@octref Thanks for merging. Indeed the issue was that this file will be updated when a new component is added/removed to the project. My local testing it was only being applied when I restarted either IDE or server.

octref:
@pi0 When you add a new component, isn't it a Vue file? Can Vetur analyze it directly or only Nuxt contain the logic to statically analyze it?

I suggest that you open a new issue that describe:
- The current limitation
- The current way you are analyzing new components and updating the JSON files

There might be other alternatives than re-reading the JSON files in `node_modules`. Vetur doesn't watch that directory since it's expensive.

octref:
I forgot to mention. If you can please update https://github.com/vuejs/vetur/blob/master/docs/framework.md as well.

pi0:
Hi @octref :) 

Docs PR: #1990

Regarding JSON, it is not inside `node_modules` but `.nuxt/vetur/tags.json` from project. Sure i will open an issue to discuss about it

99linesofcode:
@pi0 I realize this isn't your primary focus but perhaps you could take a stab at this since you also worked on the initial PR? cc: @octref

PR #1979, opened 2020-06-10
Allow analyzing Vue files in `node_modules`

### LINKED ISSUE #1127: Allow analyzing Vue files in `node_modules`

## Info

- Platform: Linux
- Vetur version: 0.16.2
- VS Code version: 1.30.2

## Problem

This is more of a feature request. I'm currently working on a Vue/TypeScript project in which we are sharing an shared, _uncompiled_ "source" package between different projects. Currently Vetur appears to avoid analyzing anything in a `node_modules` path causing the imported type to fall back on the generic shim interface definition.

I have removed `node_modules` check from here in my extension locally and confirmed the imported type is resolved correctly:

https://github.com/vuejs/vetur/blob/73a0854b9eba3311951546a1e82a6832b523acbb/server/src/modes/script/serviceHost.ts#L219

I would like to have the option to include packages in `node_modules` for analysis. For TypeScript's part I simply need to add the package to the `includes` section of `tsconfig.json` and it will work. Perhaps the option to trust the TypeScript includes would suffice?

tiagoroldao:
- Include ".vue" files in projectFiles (using ScriptKind.Deferred)
- Use project files to determine virtual file interpolation (instead of checking for `node_modules` in path)

Fixes #1127

### DIFF
--- a/server/src/services/typescriptService/serviceHost.ts
+++ b/server/src/services/typescriptService/serviceHost.ts
@@ -77,7 +77,6 @@ export function getServiceHost(
   updatedScriptRegionDocuments: LanguageModelCache<TextDocument>
 ): IServiceHost {
   patchTS(tsModule);
-  const vueSys = getVueSys(tsModule);
 
   let currentScriptDoc: TextDocument;
 
@@ -96,6 +95,8 @@ export function getServiceHost(
     `Initializing ServiceHost with ${initialProjectFiles.length} files: ${JSON.stringify(initialProjectFiles)}`
   );
   const scriptFileNameSet = new Set(initialProjectFiles);
+
+  const vueSys = getVueSys(tsModule, scriptFileNameSet);
 
   const vueVersion = inferVueVersion(tsModule, workspacePath);
   const compilerOptions = {
@@ -482,6 +483,6 @@ function getParsedConfig(tsModule: T_TypeScript, workspacePath: string) {
     /*existingOptions*/ {},
     configFilename,
     /*resolutionStack*/ undefined,
-    [{ extension: 'vue', isMixedContent: true }]
+    [{ extension: 'vue', isMixedContent: true, scriptKind: ts.ScriptKind.Deferred }]
   );
 }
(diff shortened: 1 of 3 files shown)


tiagoroldao:
> Good work. Thank you and sorry for my late response.
> 
> Is there a specific reason you are making the `ScriptKind.Deferred` change?
> I tested and there I didn't see any problem, so I'm ok for merging it.

@octref the config returned by `getParsedConfig` was not returning any vue filepaths, due to `scriptKind` not being set as Deferred (on [commandLineParser.ts](https://github.com/microsoft/TypeScript/blob/e92afacc4408065225cc7302bcfaf2c255f737d9/src/compiler/commandLineParser.ts#L2271), which calls [getSupportedExtensions](https://github.com/microsoft/TypeScript/blob/faf128de15c418d293d68e99497261cb515fc1d3/src/compiler/utilities.ts#L6356))

This was masked by the fact that the extension checked for files itself (manually ignoring `node_modules`)

tiagoroldao:
@octref on a separate note, it may make sense to set this to an even more permissive:
PS an even more permissive setting would be:
```ts
export function isVirtualVueFile(path: string, projectFiles: Set<string>) {
  return path.endsWith('.vue.ts') && 
    (!path.includes('node_modules') || projectFiles.has(path.slice(0, -'.ts'.length)));
```

octref:
> the config returned by getParsedConfig was not returning any vue filepaths, due to scriptKind not being set as Deferred (on commandLineParser.ts, which calls getSupportedExtensions)

@tiagoroldao Do you mind sending another PR that documents this inline?

> PS an even more permissive setting would be:

👍 , would be great if you can combine the changes into one PR. Being a bit looser is better, as many users don't propoerly setup jsconfig/tsconfig, so some Vue files might not be included in the TS project.

tiagoroldao:
Will do.

octref:
@tiagoroldao I added your suggestion directly since I need to make a new release, and I want to get the `isVirtualVueFile` change in.

tiagoroldao:
Eheh, I was busy making the PR, only saw this now! Feel free to close #2112 if it is now irrelevant!

Issue #1364, opened 2019-07-24
Custom tags intellisense for local files


rohit-gohri:
- [x] I have searched through existing issues
- [x] I have read through [docs](https://vuejs.github.io/vetur)
- [x] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

## Info

- Platform: Linux (Ubuntu 16.04)
- Vetur version: 0.21.1
- VS Code version: 1.36.1

## Problem

I found that there is no (documented) way to specify a `tags.json` and `attributes.json` file for custom-tags in one's own project, that are not in any dependency.

I only found https://vuejs.github.io/vetur/framework.html#custom-tags-attributes. But that is only for modules being published on npm and being used as a dependency. What about custom-tags in own project?

I tried adding a "vetur" key to my project and adding corresponding files, but it didn't pick anything up.

I think this should be trivial to implement as the major work is already done.

phiter:
This functionality also has no proper documentation whatsoever.

I could only have an idea of what is possible by looking at other framework implementations.

It shouldn't be that hard to make a table with the available options for both tags and attributes.

rchl:
I see that PR adds an example for tags.json. It would be also nice to provide an example for attributes.json.

octref:
@rchl I plan to have a more comprehensive doc for the format in by next release. I'll fix #1757 with it. Please subscribe to that issue.

PR #2018, opened 2020-07-06
feat: allow load customTags from ./package.json

pikax:
fix https://github.com/vuejs/vetur/issues/1364

Allowing to load custom tags from the package.json on the root


octref:
Sorry for my late response. Looks all good and I just made a few tweaks. I need to add docs and test as well.

octref:
Would fix #1364.

phiter:
Whoa this is amazing! I've been waiting for it for a while!

I hope this can be improved upon so you don't need to restart VSCode to see changes.

PR #1789, opened 2020-03-17
Speed up and stabilize e2e test

yoyo930021:
- [x] retry to get diagnostics until length > 0 or timeout
- [x] remove all `sleep` function and can await extension `activate`

PR #2051, opened 2020-07-23
chore: making copy:snippets run on windows

pikax:
instead of using bash converted the script to javascript


build/copy-snippets.js
```diff
@@ -0,0 +1,12 @@
+const rimraf = require('rimraf');
+const { resolve } = require('path');
```
octref, line 2: Unused import

build/copy-snippets.js
```diff
@@ -0,0 +1,12 @@
...
+const { ncp } = require('ncp');
+const { ensureDirSync } = require('fs-extra');
+
+const PDIR = 'server/dist/modes/vue';
+const DIR = 'server/dist/modes/vue/veturSnippets';
+
+ensureDirSync(PDIR);
```
octref, line 9:
```suggestion
// Create parent dir and ensure clean snippet dir
ensureDirSync(PDIR);
```

test/lsp/codeAction/basic.test.ts
```diff
@@ -17,6 +17,8 @@ describe('Should do codeAction', () => {
   });
 
   it('finds codeAction for unused import', async () => {
+    await sleep(2000); // waiting more, on windows is really slow
```
octref, line 20: Are you sure this is the cause? I think we should not add more random sleep but improve Windows integration test.

Let's not make this change just to pass CI. If it's flakey you can run it multiple times by closing/reopening.

yoyo930021: I think #1789 can stable this problem.

pikax: I'm using Windows and I can't reproduce this issues, I agree a better approach is desired

test/lsp/codeAction/basic.test.ts
```diff
@@ -43,9 +43,9 @@ async function testCodeAction(docUri: vscode.Uri, range: vscode.Range, expectedA
     range
   )) as CodeAction[];
 
-  expectedActions.forEach(eAction => {
+  expectedActions.forEach((eAction) => {
```
octref, line 46:
```suggestion
  expectedActions.forEach(eAction => {
```

octref: I removed this change. Sorry. I have just updated prettier config in master so this shouldn't happen again.

octref:
LGTM, thanks.

PR #2052, opened 2020-07-23
build(dev-deps): bump mocha to 8.0.1

pikax:


octref:
Rebased and you are probably using prettier 2 that introduced some `(x) =>` parens. I updated the prettier config in master and removed those parens.

octref:
I took the time to update server test as well.

octref:
Good job, thank you 🙌

PR #2056, opened 2020-07-24
Handle same tag name. Fix #2049

### LINKED ISSUE #2049: Component props auto-completion doesn't work when using PascalCase

## Info

- Platform: macOS
- Vetur version: 0.25.0
- VS Code version: 1.47.2

## Problem

Props auto completion works when using `<kebab-case/>` components:

Screen Shot 2020-07-22 at 12 32 13 PM

...but not with `<PascalCase/>`:

Screen Shot 2020-07-22 at 12 32 26 PM

Also, component name completion is currently also `<kebab-case/>` only:

Screen Shot 2020-07-22 at 12 34 42 PM

Would be nice to have the component name auto-completion to be based on the exact registered name.

yoyo930021:
Fixed #2049


octref:
Good work as always, thanks 👍

PR #2057, opened 2020-07-24
Add postinstall script to install all dependencies

yoyo930021:


package.json
```diff
@@ -7,6 +7,7 @@
   "version": "0.25.0",
   "publisher": "octref",
   "scripts": {
+    "postinstall": "cd server && yarn && cd ../vti && yarn && cd ../",
```
octref, line 10: I'd say just use `(cd server && yarn) && (cd vti && yarn)`. It's cleaner.

yoyo930021: Done.

yoyo930021: @octref this script can run in windows?

octref: This won't run on Windows. I suggest that we use https://github.com/mysticatea/npm-run-all

octref: You need to update https://github.com/vuejs/vetur/blob/master/.travis.yml and https://github.com/vuejs/vetur/blob/master/azure-pipelines.yml as well.

yoyo930021: I'm done.

test/lsp/features/completion/template.test.ts
```diff
@@ -1,4 +1,4 @@
-import { activateLS, showFile, sleep, FILE_LOAD_SLEEP_TIME } from '../../helper';
```
octref, line 1: I know it's tempting to do these all together, but please keep each PR focused on one single task. For 5725dc49c02b19a7e075cd4996567c322416e6cc you can open another PR. This PR should only address the postinstall script issue.

octref:
These all needs to be updated:

image

yoyo930021:
@octref I'm done.

octref:
LGTM, thanks 👍

PR #2073, opened 2020-07-30
Try adding yarn cache to CI

octref:


yoyo930021:
Add `yarn config set cache-folder '$(Pipeline.Workspace)/.yarn'`?

octref:
> Add yarn config set cache-folder '$(Pipeline.Workspace)/.yarn'?

Good point, will try.

But I'm now doubting if we should use cache, restoring .vscode-test (~200MB) takes 7 minutes:

image

octref:
I didn't realize the variable `YARN_CACHE_FOLDER` is used by `yarn` as well...Thought it's only for simplifying the bash script.

octref:
Seems much faster now:

image

Issue #1709, opened 2020-02-14
File '/Users/xxx/components/Logo.vue' is not a module 


ppzzppz:
- [ x ] I have searched through existing issues
- [ x ] I have read through [docs](https://vuejs.github.io/vetur)
- [ x ] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

## Info

- Platform: macOS
- Vetur version: https://vuejs.github.io/vetur
- VS Code version: 
```
Version: 1.42.1
Commit: c47d83b293181d9be64f27ff093689e8e7aed054
Date: 2020-02-11T14:44:27.652Z
Electron: 6.1.6
Chrome: 76.0.3809.146
Node.js: 12.4.0
V8: 7.6.303.31-electron.0
OS: Darwin x64 18.7.0
```

## Problem
error: `File '/Users/xxx/components/Logo.vue' is not a module ` in vue file

![image](https://user-images.githubusercontent.com/34941345/74534292-5dfbc780-4f6e-11ea-8a02-93cca736e162.png)

import `vue component` is ok in ts file
![image](https://user-images.githubusercontent.com/34941345/74534578-fc882880-4f6e-11ea-98e4-d28efd314a2d.png)

nuxt build also is ok 

## Reproducible Case

- create Nuxt.js + Typescript project

davelsan:
Can reproduce. Also, possible duplicate of #1716. As described in that issue, creating an empty export removes the error.

```ts
import Vue from 'vue'
export default Vue.extend({
  //
})
```

ppzzppz:
Thanks, to add `script`  resolve

<img width="574" alt="image" src="https://user-images.githubusercontent.com/34941345/75879275-12925600-5e56-11ea-8399-3b348e024b42.png">

crhistianramirez:
Looks like you can add this setting in vetur without having to create an empty export :
```json
 "vetur.experimental.templateInterpolationService": true,
```

As mentioned [in this comment](https://github.com/vuejs/vetur/issues/1187#issuecomment-489638186)

AllanOricil:
How can I fi it? It seems Im still with it.
![image](https://user-images.githubusercontent.com/55927613/84552908-be1a3480-ace8-11ea-9a39-f2ab99a821ae.png)

AllanOricil:
Hi guys. I noticed my mistake. My component was declared with .vue but because it had no implementation at all, it was not being understand by Vetur. Is it expected?

PR #2137, opened 2020-08-08
Use pure server command

yoyo930021:
1. Prepare for #1495
2. https://github.com/vuejs/vetur/pull/1704#issuecomment-668793238

### DIFF
--- a/server/src/services/vls.ts
+++ b/server/src/services/vls.ts
@@ -520,7 +520,7 @@ export class VLS {
       return;
     }
 
-    logger.logDebug(`Unknown command ${arg.command}.`);
+    logger.logInfo(`Unknown command ${arg.command}.`);
   }
 
   removeDocument(doc: TextDocument): void {
--- a/server/src/services/vls.ts
+++ b/server/src/services/vls.ts
@@ -18,7 +18,9 @@ import {
   DocumentSymbolParams,
   CodeActionParams,
   CompletionParams,
-  CompletionTriggerKind
+  CompletionTriggerKind,
+  ExecuteCommandParams,
+  ApplyWorkspaceEditRequest
 } from 'vscode-languageserver';
 import {
   ColorInformation,
@@ -36,7 +38,8 @@ import {
   TextDocumentChangeEvent,
   TextEdit,
   ColorPresentation,
-  Range
+  Range,
+  WorkspaceEdit
 } from 'vscode-languageserver-types';
 
 import { URI } from 'vscode-uri';
@@ -51,6 +54,7 @@ import { VueHTMLMode } from '../modes/template';
 import { logger } from '../log';
 import { getDefaultVLSConfig, VLSFullConfig, VLSConfig } from '../config';
 import { LanguageId } from '../embeddedSupport/embeddedSupport';
+import { Commands } from '../utils/commands';
 
 export class VLS {
   // @Todo: Remove this and DocumentContext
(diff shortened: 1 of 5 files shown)


server/src/utils/commands.ts
```diff
@@ -0,0 +1,3 @@
+export enum Commands {
+  CodeAction = 'vetur.chooseCodeAction'
```
rchl, line 2: I think that this name is not very appropriate. This command has nothing to do with code actions apart from being referenced from a code action. But commands can as well be triggered manually, without having to use code actions and in that case, that name makes no sense.

Since this command is currently used to call `getRefactorEdits`, I'd keep the old name `vetur.chooseTypeScriptRefactoring` or maybe something more in line with other servers `vetur.applyTypeScriptRefactoring`

server/src/services/vls.ts
```diff
@@ -493,6 +499,17 @@ export class VLS {
     return diagnostics;
   }
 
+  async executeCommand(arg: ExecuteCommandParams) {
+    logger.logDebug(`executeCommand: ${JSON.stringify(arg)}`);
```
rchl, line 503: I'm not sure what's the practice in this project, but I think that adding logs to log LSP communication shouldn't be necessary since there are better ways to log LSP traffic in general (The `verbose` setting in VSCode and such). I would think that `logger.logDebug` should only be used for internal things that can not be observed from LSP logs.

server/src/utils/commands.ts
yoyo930021: In #1709, I want to use it for organize imports or fix all.
CodeAction is the best name for this.

yoyo930021: The getRefactorEdits will rename when I add.

server/src/services/vls.ts
yoyo930021: You said makes sense.
Maybe we only need to log unknown commands.

server/src/utils/commands.ts
rchl: No. Not if it's a server command.
The code action is one of the ways to trigger a server command. The server commands are otherwise quite separate concepts from code actions and can be triggered without any intermediary code actions.

> I want to use it for organize imports or fix all.

Why is that a good argument for keeping the current name though?

If "organize imports" is another variant of typescript refactoring then my suggestion still makes sense.

If it will have nothing to do with typescript refactorings then there are two options:
- rename to something else that is generic but doesn't mention code actions
- create separate command for "organize imports" like `vetur.applyOrganizeImports`

rchl: And you don't need to keep reusing the same command for all sorts of semi-related actions. I'd say it's better to have separate commands.

yoyo930021: I agree about we need new name.
This part is only required in Typescript or different for language.
I think there is a need to standardize the function of this response operation.

yoyo930021: Maybe we don't change `getRefactorEdits` function and command,
We add new standardize the function and command.

rchl: I'm not sure what you are thinking now.

I think this change makes sense. It's superior to utilize standardized executeCommand over a custom command as the latter needs a client-side implementation which is against the philosophy of LSP (make once, use everywhere).

yoyo930021: Oh, I'm sorry for my English is not well.


I also unwanted about `client-side` commands.
I want to add some feature with unlike other languages.
For example, LSP don't define `Fix all CodeFix`, but Typescript have.
I want to add common functions and server commands for WorkspaceEdit.
Because the vetur supports many language, I don't want to make more functions for typescript features.

rchl: Right, so we just need a better name for the command. Maybe `vetur.applyRefactor`, `vetur.organizeImports`, `vetur.applyFixAll`. I don't think you need to put everything under the same command.

octref: I think `vetur.applyRefactorEdits` or `vetur.applyRefactorCommand` makes sense. @yoyo930021 we can rethink about standardizing when doing #1495.
And let's just put it as a constant in `javascript.ts` instead of putting it under `utils`.

yoyo930021: Ok

PR #2145, opened 2020-08-08
feat: use Vue TS plugin

znck:

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "vscode": "^1.47.0"
   },
   "dependencies": {
+    "@vuedx/typescript-vetur": "^0.2.1",
     "vscode-languageclient": "^6.1.3",
     "vscode-languageserver-types": "^3.15.1"
   },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -511,9 +512,54 @@
...
+          "type": "boolean",
+          "default": false,
+          "description": "Enable/disable automatic closing of JSX tags. Requires using TypeScript 3.0 or newer in the workspace."
         }
       }
-    }
+    },
+    "typescriptServerPlugins": [
+      {
+        "name": "@vuedx/typescript-vetur",
```
rchl, line 555: What is `@vuedx/typescript-vetur` and where is its source code located?

znck: https://github.com/znck/vue-developer-experience/blob/master/packages/typescript-vetur/src/index.ts

znck: It's TS plugin version for vetur.

/cc @octref

octref:
I'm trying renaming on https://github.com/octref/veturpack/blob/e497ab2c4000c0139d94c2dd4599fa92a4b79530/src/Test.vue#L3, for either `child` or `foo` or `bar`, but it only updates the `<template>` region.

znck:
Vue 3 compiler treats lower case tag names as elements or web components: https://github.com/vuejs/vue-next/blob/50adc014f59a7f385c836014abb059ec72c52239/packages/compiler-core/src/parse.ts#L483-L487 so renaming `child` won't update the `Child` component registered in the `components` option.

Variable names `foo` and `bar` we not working due to a bug that made a false assumption that every component is using `<script setup>`.

yoyo930021:
@znck I think we can add a option for ts plugin in Vetur.
You can update this PR if you have time.
I will merged it and add option for it.

znck:
Let's wait couple more weeks, I'll try to make it a bit more stable.

yoyo930021:
@znck Is the time ripe?

znck:
I need to make one more change for 2.x support, using `defineCompoment` types from composition API plugin.

yoyo930021:
> I need to make one more change for 2.x support, using `defineCompoment` types from composition API plugin.

Is `defineCompoment` types better than `Vue.extends` in Vue 2?
If user project doesn't have `@vue/composition-api`?

PR #2161, opened 2020-08-14
Fix #1484

### LINKED ISSUE #1484: "Go to definition" for methods/computeds does not work in the template - Unable to open '[somefile].vue': resource is not available.

## Info

- Platform: Windows 10 
- Vetur version: 0.22.6
- VS Code version: 1.39.2

## Problem
Running "Go to definition" on a method/computed name inside the template (with `vetur.experimental.templateInterpolationService` enabled) throws an error:
`Unable to open 'HelloWorld.vue': resource is not available.`

I tested this in Veturpack and the default (babel, eslint) Vue CLI 4 project.

More verbose message from `Log (Window)` in Output panel:
```
[2019-10-31 22:17:30.846] [renderer1] [error] resource is not available: Error: resource is not available
    at v.createReferencedObject (file:///C:/Users/atagisow/AppData/Local/Programs/Microsoft VS Code/resources/app/out/vs/workbench/workbench.desktop.main.js:4888:687)
    at v.createReferencedObject (file:///C:/Users/atagisow/AppData/Local/Programs/Microsoft VS Code/resources/app/out/vs/workbench/workbench.desktop.main.js:4888:652)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

vetur

When the above gif was made:
* Vetur was the only enabled extension
* The contents of settings.json was only `"vetur.experimental.templateInterpolationService": true`

## Reproducible Case
*the gif I uploaded is made on the Vue CLI 4 template, here I'll show how to reproduce it in Veturpack*

1. Clone Veturpack
2. In VSCode go to `veturpack\client\components\Counter.vue`
3. Right click `count` on line 8 and press "Go to definition"

You'll get the `Unable to open 'Counter.vue': resource is not available.` error

I'm not sure if what I'm describing is even a feature of Vetur? I'm going by what I've read in https://vuejs.github.io/vetur/interpolation.html#generic-language-features -> :

`Currently diagnostics, hover, jump to definition and find references are implemented in this way:`

So I'm assuming I should be able to do what I'm trying to do.

This issue is different from #707 because it's related to methods/computeds whereas 707 is related to Go Tos in imports

cereschen:


cereschen:
still have bug

cereschen:
For example
 <span v-for="item in items" v-text="item" />

Find definitions for item is not work well
Even report an error
But I can't fix this right now

octref:
Maybe you are running into #2167. Otherwise this looks good to me and makes the URI handlign consistent with `javascript.ts`.

Since JS jump to definition is working for people so this should work across macOS / Windows as well.

Thanks for your contribution 👍

cereschen:
> Maybe you are running into #2167. Otherwise this looks good to me and makes the URI handlign consistent with `javascript.ts`.
> 
> Since JS jump to definition is working for people so this should work across macOS / Windows as well.
> 
> Thanks for your contribution 👍

I mean, the variables created in the v-for directive can't find the definition correctly
But it seems that this is a problem that can be ignored
Vetur is a great extension, and I hope to help improve the user experience

Issue #2141, opened 2020-08-08
Prop with `required: false` or default value should not trigger prop validation error when not provided


octref:
From https://github.com/vuejs/vetur/issues/2135#issuecomment-670841767

octref:
- Use `DiagnosticSeverity.Error` for passing an undeclared prop
- When not passing a required prop
  - If in array form, show a warning with `DiagnosticSeverity.Warning` (as all props are optional)
  - If in object form, only show error if required props are not passed. For example, if a component requires `foo` and takes an `optional` bar, not passing `bar` should not generate an error.

yoyo930021:
When no prop:
```
// case 1 -> warning
props: [foo, bar]
// case 2 -> warning
props: {
  foo: String,
  bar: String,
}
// case 3 -> error
props: {
  foo: {
    type: String,
    required: true
  }
}
// case 4 -> No any errors or warning
props: {
  foo: {
    type: String,
    required: false
  },
  bar: {
    type: String,
    default: ''
  }
}
```

madarame-haruto20-fixer:
Hi,

When using `v-bind="<Prop Object>"`, this will show `misses props` warning.

![image](https://user-images.githubusercontent.com/64626169/92189855-a6db8500-ee9a-11ea-8776-bbcbf1199024.png)

Is there any solution to fix it?

PR #1323, opened 2019-06-07
Add support for analyzing vue-class-component and vue-property-decorator

yoyo930021:
Fixed #1105.
Fixed #864.

I want to support vue-property-decorator.
Reason:
1. Vue 3.0 isn't built-in class component. [rfc](https://github.com/vuejs/rfcs/pull/17)
2. vue-property-decorator is built-in vue-cli 3 typescript plugin.
3. [vue-typescript](https://github.com/itsFrank/vue-typescript)  project latest update was three years ago.

I'm not add test.
I will try to add.

PR #2164, opened 2020-08-14
Fix PropSync and Model decorators for vue property decorator

yoyo930021:
For #1323 

I find some problems when doing #2141 
I write some wrong logics when use `PropSync` and `Model` decorators.

ref:
- https://github.com/kaorun343/vue-property-decorator#-propsyncpropname-string-options-propoptions--constructor--constructor---decorator
- https://github.com/kaorun343/vue-property-decorator#-modelevent-string-options-propoptions--constructor--constructor---decorator

### DIFF
--- a/server/src/modes/script/componentInfo.ts
+++ b/server/src/modes/script/componentInfo.ts
@@ -113,10 +113,24 @@ function getProps(tsModule: T_TypeScript, defaultExportType: ts.Type, checker: t
       return undefined;
     }
 
-    return propsSymbols.map(prop => {
+    return propsSymbols.map(propSymbol => {
+      const prop = propSymbol.valueDeclaration as ts.PropertyDeclaration;
+      const decoratorExpr = prop.decorators?.find(decorator =>
+        tsModule.isCallExpression(decorator.expression)
+          ? propDecoratorNames.includes(decorator.expression.expression.getText())
+          : false
+      )?.expression as ts.CallExpression;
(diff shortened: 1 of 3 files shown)


server/src/modes/script/componentInfo.ts
```diff
@@ -113,10 +113,24 @@ function getProps(tsModule: T_TypeScript, defaultExportType: ts.Type, checker: t
...
+      const prop = propSymbol.valueDeclaration as ts.PropertyDeclaration;
+      const decoratorExpr = prop.decorators?.find(decorator =>
+        tsModule.isCallExpression(decorator.expression)
+          ? propDecoratorNames.includes(decorator.expression.expression.getText())
+          : false
+      )?.expression as ts.CallExpression;
+      const decoratorName = decoratorExpr.expression.getText();
+      const args = decoratorExpr.arguments;
+
+      const firstNode = args[0];
```
octref, line 126:
```suggestion
      const firstNode = decoratorExpr.arguments[0];
```

octref (approved):
Good work 👍 Just a small nitpick. No need to create only-use-once variables.

octref:
I'll just commit the small change and merge.

server/src/modes/script/componentInfo.ts
yoyo930021: Oh, It's just use in #2166

PR #2165, opened 2020-08-14
Remove npm in package.json and add yarn command in docs

yoyo930021:


octref:
Thanks for fixing docs/bugs till late night 🍻

PR #2176, opened 2020-08-16
Extend HTML mode suggestions for child components (props & events)

triforcely:
This PR isn't related to specific issue but it fixes some things that personally annoy me in my daily work with vue & vetur. I'm opening it as a draft to gather feedback.

Short log of what I've done:
- events from SFC child components will be suggested in templates with type of event handling function (according to vue-property-decorator spec)
- extended component/child component parsing logic to retrieve their locations in files (making referencing child component files possible, e.g. for go to definition)
- "hover" and "go to definition" for SFC child component props and events

I believe that a lot of this functionality could be extended to the object component syntax too.

Now some examples:
<details>
  <summary>Hover info on child component event</summary>

  image

</details>

<details>
  <summary>Hover info on child component property:</summary>

image

</details>

<details>
  <summary>Suggestions for events</summary>

image
</details>

<details>
  <summary>Go to definition (GIF)</summary>

image

</details>


cereschen:
My own extension implement these features, but Vetur is a strict project and there are too many things to consider

triforcely:
I agree but I still believe it is worth pursuing good support, on par with other more mature frontend frameworks. I can put few hours a week as long as I get some guidance from core maintainers. There are thousands of developers that rely on Vetur and some features could save years of time.

cereschen:
It would be better if Vetur could support third-party plugins, as frequent transform to AST can cause performance problems

triforcely:
Yeah, that is a good idea - users could install support for only needed languages/types.

server/src/modes/script/componentInfo.ts
```diff
@@ -375,6 +416,56 @@ function getMethods(
...
+          position: getRangeFromNode(doc, d.expression)
+        };
+
+        return methodInfo;
+      });
+    });
+  }
+
+  function getObjectEvents(type: ts.Type, node: ts.Node, checker: ts.TypeChecker) {
+    // TODO: not yet supported, could traverse AST in search for "$emit" calls.
```
yoyo930021, line 465: We cant do about this.
The `$emit` event name can dynamic.

Maybe we can use new `emits` options.
https://github.com/vuejs/rfcs/blob/3f9eebb0162fb18f8f9f8f22523ae4598d198740/active-rfcs/0030-emits-option.md

https://github.com/vuejs/rfcs/blob/3f9eebb0162fb18f8f9f8f22523ae4598d198740/active-rfcs/0030-emits-option.md

triforcely: Maybe the type can be inferred using TypeChecker?

yoyo930021: Example:
```javascript
export default {
  methods: {
    open(event) {
      this.$emit(event);
  }
}
```

triforcely: Yeah, this will result in some "any" type being detected. But I'd say that the responsibility for having correct types lies on the developer, so for having correct type it would have to be written as:
```js
export default {
  methods: {
    open(event: EventType) {
      this.$emit(event);
    }
  }
}
```
Which is less verbose than proposed RFC. I'll try to implement PoC.

triforcely: Hey @yoyo930021, I implemented quick version for getting events based on the TS TypeChecker:
image

This works when event name is string literal. I can't think about use case where my event name would be dynamic.

server/src/modes/template/services/htmlCompletion.ts
```diff
@@ -138,30 +138,26 @@ export function doComplete(
   }
 
   function collectAttributeNameSuggestions(nameStart: number, nameEnd: number = offset): CompletionList {
-    const execArray = /^[:@]/.exec(scanner.getTokenText());
-    const filterPrefix = execArray ? execArray[0] : '';
-    const start = filterPrefix ? nameStart + 1 : nameStart;
```
triforcely, line 143: This caused props to be completed as "::propName=" instead of ":propName=" when ":" was already typed

server/src/modes/script/componentInfo.ts
philippevezina: I agree with @yoyo930021 that if events are specified in the new `emits` options they should be displayed as a suggestion. Hoping to see this feature soon!

octref:
Hey @triforcely, sorry I haven't had time for reviewing.
It would be easier for me to review if you can split the PR into smaller chunks, each one with a single fix/feature.
Do you mind keeping this open so I can get back to it at some point later?

triforcely:
@octref There is a lot of ways to do the same thing in vue, so it will be hard to fully implement support for it that works with vue 2 and 3 unless there are strict requirements defined, so I can make sure that I don't forget anything. I can implement it but my vue knowledge is limited to what I use in my daily work (Vue2, TS and property decorator).

triforcely:
@octref Maybe we could have an issue where we track missing functionality of the intellisense/diagnostic features so we could discuss these kinds of requirements there?

yoyo930021:
> @octref Maybe we could have an issue where we track missing functionality of the intellisense/diagnostic features so we could discuss these kinds of requirements there?

Ha, Good idea, You can open a issue.👍

Issue #2179, opened 2020-08-19
Component Data json format


michaeltintiuc:
- [x] I have searched through existing issues
- [x] I have read through [docs](https://vuejs.github.io/vetur)
- [x] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

I've read the Component Data docs and have looked through 3rd party implementations of attributes.json and tags.json, but I couldn't find a way to allow for auto-completion and suggestion of custom events, I've tried prefixing props with `v-on:` and `@` but it's not quite the same when you type `@` and it brings up a list of all available events to listen on.

rchl:
I think the current format is very limited. See https://github.com/vuejs/vetur/issues/2090 for something more complete.

octref:
I'll close after adding some doc.

michaeltintiuc:
Thanks @octref , just to clarify, an event type would allow for auto-completion and suggestions with `v-on` and `@`?

octref:
@michaeltintiuc Yes, sorry for the lack of documentation now...Take a look at https://github.com/vuejs/vetur/pull/2180/commits/1ccb6ba88898f003f484842be66c70c7e9c37d9d
and the setup here: https://github.com/vuejs/vetur/blob/813a3b52157e7a79a17383a320eac88f85b2050f/test/componentData/fixture/package.json#L10-L13

I'll make a release this week.

octref:
Cool modeling btw 🎩

michaeltintiuc:
Thanks a lot @octref !

michaeltintiuc:
Is this scheduled for an upcoming release? If there's anything I can help with I'd be happy to

michaeltintiuc:
I was also wondering if there's any reason to support `events` prop in `tags.json` same it's done for `attributes`, i.e.:
```json
{
  "foo": {
    "description": "Foo tag",
    "attributes": ["bar"],
    "events": ["baz"]
  }
}
```

Issue #2173, opened 2020-08-16
Emmet didn't work after curly parentheses


pixieaka:
```
<div class="home">
    {{ user.name }}
    // Emmet didn't work here in this case when i write `p` it didn't auto complete with emmet extention
</div>
```

octref:
Fixed in #2180.

PR #2180, opened 2020-08-20
Fix #1678 #2173 #2179

### LINKED ISSUE #1678: No CSS completions in `style` attribute inside `<template>` region
## Info
- Platform: macOS
- Vetur version:
- VS Code version:

## Problem

"files.associations": {
        "*.vue": "vue"
  }
then:
image

style snippet not work

"files.associations": {
        "*.vue": "html"
  }
then:
image

Vetur does not work

I hope to have code hints for inline styles, and also use Vetur.
Can this be achieved?

image
image

cereschen:
Now emmet works normally after interpolation:

image

Inline style emmet auto-complete support:

image

Custom event support:

image
image

### DIFF
--- a/server/src/modes/template/services/htmlCompletion.ts
+++ b/server/src/modes/template/services/htmlCompletion.ts
@@ -286,7 +286,8 @@ export function doComplete(
       case TokenType.AttributeValue:
         if (scanner.getTokenOffset() <= offset && offset <= scanner.getTokenEnd()) {
           if (currentAttributeName === 'style') {
-            return emmet.doComplete(document, position, 'css', emmetConfig);
+            const emmetCompletions = emmet.doComplete(document, position, 'css', emmetConfig);
+            return emmetCompletions || NULL_COMPLETION;
           } else {
             return collectAttributeValueSuggestions(
               currentAttributeName,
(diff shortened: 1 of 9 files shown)


server/src/modes/template/services/htmlCompletion.ts
```diff
@@ -285,11 +285,15 @@ export function doComplete(
         break;
       case TokenType.AttributeValue:
         if (scanner.getTokenOffset() <= offset && offset <= scanner.getTokenEnd()) {
-          return collectAttributeValueSuggestions(
-            currentAttributeName,
-            scanner.getTokenOffset(),
-            scanner.getTokenEnd()
-          );
+          if (currentAttributeName === 'style') {
+            return emmet.doComplete(document, position, 'css', emmetConfig);
```
octref, line 289: During my test this would return `undefined` for such cases `<div style="|"></div>`. Need to handle that.
I think the result should be a combination of emmet/collectAttributeValueSuggestions.

octref:
I found for `poa`, you need to manually trigger completion. This isn't ideal, but I haven't figured why that's happening. Other changes are good. Thank you 💯 🙏

octref:
However I wouldn't say that's the correct way of setting up Component Data...Let me write the format down first.

cereschen:
> However I wouldn't say that's the correct way of setting up Component Data...Let me write the format down first.

It's okay. I made my own decision 😅

octref:
I added some integration test. Please familiarize yourself with how to add them for future PRs:

https://github.com/vuejs/vetur/wiki/Pull-Request-Guidance#integration-test

PR #2191, opened 2020-08-21
Add showing #slot attrs in symbols

### LINKED ISSUE #2169: Implement more detailed symbol info for tags with slot-related attributes

## Feature Request

1. Show slots in vscode symbol browser:
obraz

2. Show `slot[name="slotname"]` in symbol search instead of just `slot` in symbol search

### Things I'm not sure about:
* Should [old slot & slot-scope syntax](https://vuejs.org/v2/guide/components-slots.html#Named-Slots-with-the-slot-Attribute) be supported?
* Displaying the slot like `div#id.class #slotname` is weird as two attrs use the `#` symbol (which is normally reserved for id). Maybe `div#id.class[#slotname]` would be better? VSCode never returns attr info for symbols in .html files though (eg. `<input type='number'>` still returns symbol info like `input` rather than `input[type='number']`
*  Should slot syntax be unified in symbols? eg. should symbol search always show `template[#slot]` even though the user explicitly wrote `<template v-slot:slot>` in code?

### Rationale
Consider [this case](https://github.com/vuetifyjs/vuetify/blob/master/packages/docs/src/examples/data-tables/intermediate/slots.vue) where you have a large component with many slots. You may want to quickly jump between slots using the symbol search (CTRL+SHIFT+O).

3nuc:
fixes #2169
obraz


octref (changes requested):
Can you fix the failing tests and add either some server unit tests or integration tests? Thanks 👍 

And when `slot` is not present, I don't think it should generate the extra `[]`.

PR #2211, opened 2020-08-28
chore: use Vue's theme color for badges

antfu:
The original "green" is a bit too saturated and having less contrast with the white text in my opinion. Not sure if you like the change though.

Using the Vue's color #42b883. Now it looks like this:

image

Another idea might be to use the other color #35495e together, which is not part of this PR. If you like it, I can change to it for sure :)
image


octref (changes requested):
> Another idea might be to use the other color #35495e together

Go for it. Thanks 👍

antfu:
Done https://github.com/vuejs/vetur/blob/0610266502ecf607da1455240bbbfc8003dd015b/README.md

octref (approved):

PR #2241, opened 2020-09-04
Fix corner case and add v-model support when prop validation

### LINKED ISSUE #2236: The props validator can't handle about `v-bind`, `:[key]` and `v-model`

## Feature Request
The `v-bind` is hard to handle.
Maybe we can ignore it.

The `:[key]` is also hard to handle.

The `v-model` can read AST to handle.

### LINKED ISSUE #2237: Missing props warning when using .sync modifier

## Info

- Platform: macOS
- Vetur version: 0.27.1
- VS Code version: 1.48.2

## Problem

image

yoyo930021:
Fixed #2236
Fixed #2237

### DIFF
--- a/server/src/modes/template/services/vuePropValidation.ts
+++ b/server/src/modes/template/services/vuePropValidation.ts
@@ -49,7 +49,10 @@ function generateDiagnostic(n: Node, definedProps: PropInfo[], document: TextDoc
   const seenProps = n.attributeNames.map(attr => {
     return {
       name: attr,
-      normalized: normalizeHtmlAttributeNameToKebabCase(attr)
+      normalized: normalizeHtmlAttributeNameToKebabCase(
+        attr,
+        definedProps.find(prop => prop.defaultModel)?.name ?? 'value'
+      )
     };
   });
 
@@ -82,8 +85,18 @@ function generateDiagnostic(n: Node, definedProps: PropInfo[], document: TextDoc
   };
 }
 
-function normalizeHtmlAttributeNameToKebabCase(attr: string) {
+function normalizeHtmlAttributeNameToKebabCase(attr: string, defaultModel: string) {
   let result = attr;
+
+  // v-model.trim
+  if (!result.startsWith('v-model:') && result.startsWith('v-model')) {
+    return kebabCase(defaultModel);
+  }
+
+  // Allow `v-model:prop` in vue 3
+  if (result.startsWith('v-model:')) {
+    result = attr.slice('v-model:'.length);
+  }
 
   if (result.startsWith('v-bind:')) {
     result = attr.slice('v-bind:'.length);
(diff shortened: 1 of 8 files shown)


server/src/modes/template/services/vuePropValidation.ts
```diff
@@ -77,15 +85,30 @@ function generateDiagnostic(n: Node, definedProps: PropInfo[], document: TextDoc
   };
 }
 
-function normalizeHtmlAttributeNameToKebabCase(attr: string) {
+function normalizeHtmlAttributeNameToKebabCase(attr: string, defaultModel: string) {
```
octref, line 88: If this is `defaultModel`, don't name the othe attribute as `defaultModel`. It would confuse the reader.

octref (changes requested):

server/src/modes/template/services/vuePropValidation.ts
yoyo930021: `defaultModelAttr`?

octref: I think `boundToModel` makes more sense, as in `a prop that is bound to model`.

octref: And this one could be `modelProp`, as you are trying to figure out which prop is bound to model and passing its name in.

server/src/modes/template/services/vuePropValidation.ts
```diff
@@ -41,10 +41,18 @@ function traverseNodes(nodes: Node[], f: (n: Node) => any) {
...
+    return undefined;
+  }
+
   const seenProps = n.attributeNames.map(attr => {
     return {
       name: attr,
-      normalized: normalizeHtmlAttributeNameToKebabCase(attr)
+      normalized: normalizeHtmlAttributeNameToKebabCase(
+        attr,
+        definedProps.find(prop => prop.defaultModel)?.name ?? 'value'
```
octref, line 54: I would write my logic here:

- If it's `v-model`, create a function such as `normalizeModelAttr` that takes in the `definedProps` and find it's normalized prop. You can make this function use `normalizeHtmlAttributeNameToKebabCase`.
- If it's not, use `normalizeHtmlAttributeNameToKebabCase` directly

Currently this function is overloaded with too many things. Make one function do one thing is better (and then compose them).

server/src/modes/template/services/vuePropValidation.ts
octref: See the other comment — I'd do the removing `.sync` logic here (or removing any modifiers), but I'd do the model part in another function: https://github.com/vuejs/vetur/pull/2241/files#r484022087

octref (changes requested):

octref:
Sorry I didn't wait. I'm planning to make a release soon to address the folidng issue and this issue, so I made the changes myself.

server/src/modes/template/services/vuePropValidation.ts
yoyo930021: Your variable name is really better.
I've learned, thanks.

server/src/modes/template/services/vuePropValidation.ts
yoyo930021: I hesitated when I wrote this function,
but I'm writing this anyway.

I will send a PR to refactor it.

Issue #1707, opened 2020-02-13
Intellisense not always working in computed properties


Soresu:
- [ x] I have searched through existing issues
- [ x] I have read through [docs](https://vuejs.github.io/vetur)
- [ x] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

## Info

- Platform: Win 10
- Vetur version: 0.23.0
- VS Code version: 1.42.1

## Problem

Intellisense is not working when the computed property returns a reference from this, and the component has props, even its empty. I belive this is the same as #1509 , but the return part is not visible in the picture.

## Reproducible Case

You can check the example repo, I've changed [Counter.vue](https://github.com/Soresu/veturpack/blob/master/client/components/Counter.vue) file

![image](https://user-images.githubusercontent.com/9396091/74486248-d3a75b00-4ebc-11ea-9caf-1324bf5e5446.png)

nekosaur:
This might be platform specific? This is on OS X

<img width="823" alt="Screenshot 2020-02-17 at 20 25 41" src="https://user-images.githubusercontent.com/623137/74681468-c68fb200-51c3-11ea-886d-6396f08f84cb.png">

But on W10 I got the same result as you

alexandrmiee:
I reproduce this bug on
Win10, Ubuntu
Vetur 0.24.0
Soleved by set @returns option to all computed props
![image](https://user-images.githubusercontent.com/8451506/76674867-f1073c00-65c4-11ea-9346-be64ba35d05b.png)

F0rsaken:
It's connected with this https://vuejs.org/v2/guide/typescript.html#Annotating-Return-Types
and issue with circular dependencies in TypeScript >= 3.4.x

Even if you don't use TS, Vetur does and not annotating return type of computed kind of breaks `this` context.

cerino-ligutom:
Getting this as well. What's odd is when the `return` keyword is there, it breaks:
![image](https://user-images.githubusercontent.com/6721822/77260317-0d406280-6cc2-11ea-8db4-38942a41b2c5.png)

When not:
![image](https://user-images.githubusercontent.com/6721822/77260322-19c4bb00-6cc2-11ea-98a3-92ca06819aea.png)

F0rsaken:
@cerino-ligutom when the `return` is used, TypeScript wants to infer the return type of the computed `test` that depends on `this`, but `this` depends on also computed `test`, but `test`....

And we go into recursive loop....

[...]

PR #2250, opened 2020-09-04
Add mention about computed properties in the FAQ

Uninen:
Added some docs as per earlier discussions here: https://github.com/vuejs/vetur/issues/1707

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -66,9 +66,10 @@ For 2, try these methods:
     }
   }
   ```
-## Intellisense and computed properties
 
-There is an upstream issue with TypeScript where TypeScript cannot infer `this` type correctly in computed properties and cannot provide correct auto-completion and type-checking. Read more about the issue and workaround here: https://github.com/vuejs/vetur/issues/1707#issuecomment-686851677
+## Template errors in components written in Javascript
+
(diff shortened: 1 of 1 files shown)


rchl:
Personally I think the issue is more generic than stated in that text.

While there might have been some regression in Typescript that made it infer less often, it was never able to always infer everything properly in a JS codebase. There were always cases where annotations had to be added to make things work. Whether it's for computed values, method, or other top-level options (like `head` in Nuxt, for example).

With that said, it might be good enough but I'd make it as generic as possible. Otherwise, people will be like, "I don't have computed properties in my component", and think that issue doesn't apply to them.

My suggestion would be something more like:

## Template errors in components written in Javascript

If you have `vetur.validation.interpolation` enabled and getting seemingly spurious errors in a component, it might be due to Typescript not being able to infer types properly. In Javascript-based code base, Typescript does a lot of guessing to infer types properly but it's not always able to do that automatically. You might have to add type annotations manually in the script section. Read more about the issue here: https://github.com/vuejs/vetur/issues/1707#issuecomment-686851677. Also, check [how to add JSDoc annotations](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html).

Uninen:
Thanks for the suggestion, this is much better than my original. Changed the patch accordingly.

octref (approved):

octref:
Thanks!

PR #2297, opened 2020-09-21
Performance optimization: merge languageService

cereschen:
Use a Set to save filenames, it's can save about 100ms.

 The only difference between templateLanguageService and  jsLanguageService  is getCompletionsAtPosition.

But in every time doValidate and others  takes twice as long,merge them will greatly reduce waiting times

Now I only do doValidate(the function which wasting a lot of time saving files) part, if you pass the test, you can modify the other parts


cereschen:
Given that this will require a lot of code changes, I'll just give the idea...

Issue #2305, opened 2020-09-22
VTI prints empty filenames if the only error is an eslint-plugin-vue error


andrewisaburden:
- [x] I have searched through existing issues
- [x] I have read through [docs](https://vuejs.github.io/vetur)
- [x] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

## Reproducible Case

Create a Vue file with an eslint-plugin-vue error and no Vetur errors.
Run `vti diagnostics`
The filename will be printed with no error or warning below it

I'll fix it and make a PR, if that's OK.

PR #2306, opened 2020-09-22
2305 - Fix VTI printing empty filenames

andrewisaburden:
This was due to eslint-plugin-vue errors being ignored

Fixes https://github.com/vuejs/vetur/issues/2305

### DIFF
--- a/vti/src/cli.ts
+++ b/vti/src/cli.ts
@@ -103,7 +103,9 @@ async function getDiagnostics(workspaceUri: URI) {
         uri: URI.file(absFilePath).toString()
       })) as Diagnostic[];
       if (res.length > 0) {
-        console.log(`${chalk.green('File')} : ${chalk.green(absFilePath)}`);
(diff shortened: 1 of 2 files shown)


vti/src/cli.ts
```diff
@@ -103,7 +103,9 @@ async function getDiagnostics(workspaceUri: URI) {
         uri: URI.file(absFilePath).toString()
       })) as Diagnostic[];
       if (res.length > 0) {
-        console.log(`${chalk.green('File')} : ${chalk.green(absFilePath)}`);
+        if (res.filter(r => r.source !== 'eslint-plugin-vue').length > 0) {
+          console.log(`${chalk.green('File')} : ${chalk.green(absFilePath)}`);
+        }
```
octref, line 108: I'd say just filter out the errors in `res`, and then check `res.length`.

andrewisaburden: Like so?

octref: Exactly, thanks 👍

octref (approved):

Issue #2192, opened 2020-08-23
Sometimes Vetur is still very slow


BSG-75:
- [X] I have searched through existing issues
- [X] I have read through [docs](https://vuejs.github.io/vetur)
- [X] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

## Info

- Platform: Windows
- Vetur version: 0.26.1
- VS Code version: 1.48.1

## Problem

I've seen related issues like https://github.com/vuejs/vetur/issues/1272 talking about calls to `synchronizeHostData`. I did not dig in too much, but I found that when validating sources and templates, calls such as `getSemanticDiagnostics` may call `synchronizeHostData` internally ([link](https://github.com/microsoft/TypeScript/blob/97a072926ffe2cb870732295fd3f0608e9a4e768/src/services/services.ts#L1504)), and by a rudimentary profiling it looks like each validation will took ~300ms on my computer. When I'm writing code, js and template will be validated separately, so they might take something like 500 or 600ms in total.
I'm not sure if that's the cause of the issue since I didn't do an accurate profiling, but probably it will help.
https://github.com/vuejs/vetur/blob/53ee52746b1f26f44391c3764bc7bd06d4ddaa46/server/src/modes/template/interpolationMode.ts#L72
https://github.com/vuejs/vetur/blob/2e32351514b01f3bec9478c0e55217e21b88d8ae/server/src/modes/script/javascript.ts#L118

octref:
Can you attach a profile? If not, can you validate your hypothesis by
- cloning
- `return []` eagerly here: https://github.com/vuejs/vetur/blob/a6ce17e19c4c7360896a0c9e6ce0ca8021036bca/server/src/modes/script/javascript.ts#L110
- compiling the server
- test by configuring `vetur.dev.vlsPath` to absolute path of your local modified `/server`

BSG-75:
Tried to `return []` eagerly in both `javascript.ts` and `interpolationMode.ts`. Vetur is still quite slow, but I can feel that the speed of suggestions (for example when I type `this.` and wait for the suggestions of the component data members / props / methods) has increased significantly.
[CPU-20200827T000556.534Z.cpuprofile.txt](https://github.com/vuejs/vetur/files/5133234/CPU-20200827T000556.534Z.cpuprofile.txt)

yoyo930021:
I've studied this side.
I find that typescript language service rebuilds program every time.
[...]

Issue #2163, opened 2020-08-14
Intellisense in latest Vetur is painfully slow


kjeksfjes:
- [x] I have searched through existing issues
- [x] I have read through [docs](https://vuejs.github.io/vetur)
- [x] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

## Info

- Platform: macOS
- Vetur version: 0.26.1
- VS Code version: 1.48.0

## Problem

Intellisense, e.g. when typing `import whatever from './'` i takes around 10 seconds for suggestion to pop up.

I've tried downgrading VS Code to no avail. The only thing that seems to work is downgrading Vetur to 0.25.0.

## Reproducible Case

- Use latest Vetur on latest VS Code. Use something like Path Intellisense (doesn't matter) and try to import something by typing `import whatever from './'`.
- Expected result: Intellisense suggestion pops up right away.
- Actual result: After 10 seconds or so intellisense suggestion pops up.

yoyo930021:
Can you profile it?
https://github.com/vuejs/vetur/blob/master/.github/PERF_ISSUE.md

3nuc:
Having similar issue but with TypeScript intellisense of instance properties in <script> or <template> tags.

What I did:
1. Open https://github.com/tokyo-metropolitan-gov/covid19/blob/development/components/PositiveRateMixedChart.vue (in vscode after cloning repo and npm install, not on web)
1. Go to <template>
2. Write down `v-text="displayOptionHeader."` (notice dot at end) in some random tag
3. **Pressed start CPU profiling in Chrome**
4. Return to vscode after '.' and press ctrl+space
5. Wait for intellisense box to appear with suggestions based on TS interface
6. **Stop CPU profiling*
[covid-19-ctrl-space-in-template.zip](https://github.com/vuejs/vetur/files/5075780/covid-19-ctrl-space-in-template.zip)

I mainly experience it at work in a vue-property-decorator TS project but I don't wanna share the profile from there

IWANABETHATGUY:
Actually , most language feature of vetur come from `typescript`, so the bottleneck also came from typescript, i am not sure if vetur using the typescript incremental parsing.

3nuc:
The difference between editing a `.ts` file and `.vue` lang="ts" file in the same workspace is very noticeable. IntelliSense in `.ts` is much faster, though I use the TS nightly extension. Not sure if others share the same experience

octref:
[...]

PR #2328, opened 2020-09-25
Reduce recreate ts program when no need for ts perf

yoyo930021:
Improve a part of #2192, #2163 

This change reduce VTI run time by half.

## gtime result
old:
834.59 user 29.64 system 11:21.12 elapsed 126% CPU (0avgtext+0avgdata 1409020maxresident)k

new:
422.72 user 12.04 system 6:04.27 elapsed 119% CPU (0avgtext+0avgdata 1221284maxresident)k

but this PR is dangerous, I will observe for a while in local.


yoyo930021:
I think it's great when short time test.
I will observe it more time.

yoyo930021:
Ref: https://github.com/vuejs/vetur/issues/2192#issuecomment-699864856

octref (approved):
LGTM, thanks for finding this!

PR #2337, opened 2020-09-27
Add version command

andrewisaburden:
Add --version command

Usage:
`vti --version` or `vti -v`

Outputs
`v0.0.13`


andrewisaburden:
In the future, if more commands are added then we can consider using [Optionator](https://github.com/gkz/optionator) or [yargs](https://github.com/yargs/yargs).

octref (approved):
I agree in the future we can add flags and whatnot, but for now let's keep it simple — just `vti diagnostics/version/format` etc.

The change is great. Thanks for contributing! 👍

PR #2353, opened 2020-10-06
Fix no complete literal string union

### LINKED ISSUE #2300: Literal String Union Autocomplete now working inside .vue file

## Info

- Platform: Win10 Home
- Vetur version: 0.27.3
- VS Code version: 1.49.1

## Problem
Autocomplete for string literal union type not working inside **.vue** file (with `<script lang="ts">` ). But working just fine inside **.ts** file.

Working nice in **test.ts**
working

No autocomplete, only information in **test.vue**
not_working

## Reproducible Case
**Declare any kind of type with string literal union type.**
Expected: Give autocomplete just as in a .ts file.
Reality: No autocomplete in **.vue** file

**Note**: I'm not really sure where to post this: vue, vetur or vscode. And i also not really sure whether this should be a bug report or feature request. But since in the vetur doc they said that 

> The IntelliSense should be almost the same to the IntelliSense in js/ts files 
So maybe this is the right place to post this.

yoyo930021:
Fixed #2300

### DIFF
--- a/server/src/modes/script/javascript.ts
+++ b/server/src/modes/script/javascript.ts
@@ -151,6 +151,7 @@ export async function getJavascriptMode(
         return { isIncomplete: false, items: [] };
       }
       const completions = service.getCompletionsAtPosition(fileFsPath, offset, {
+        triggerCharacter: getTsTriggerCharacter(triggerChar),
         includeCompletionsWithInsertText: true,
         includeCompletionsForModuleExports: _.get(config, ['vetur', 'completion', 'autoImport'])
       });
@@ -758,6 +759,13 @@ function getFormatCodeSettings(config: any): ts.FormatCodeSettings {
     indentSize: config.vetur.format.options.tabSize,
     convertTabsToSpaces: !config.vetur.format.options.useTabs
   };
+}
+
+function getTsTriggerCharacter(triggerChar: string) {
+  const legalChars = ['@', '#', '.', '"', "'", '`', '/', '<'] as const;
+  type LegalChars = typeof legalChars[number];
+  if (legalChars.includes(triggerChar as LegalChars)) return triggerChar as LegalChars;
+  return undefined;
 }
 
 function convertCodeAction(
--- a/server/src/modes/script/javascript.ts
+++ b/server/src/modes/script/javascript.ts
@@ -764,7 +764,9 @@ function getFormatCodeSettings(config: any): ts.FormatCodeSettings {
 function getTsTriggerCharacter(triggerChar: string) {
   const legalChars = ['@', '#', '.', '"', "'", '`', '/', '<'] as const;
   type LegalChars = typeof legalChars[number];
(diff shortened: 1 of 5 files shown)


yoyo930021:
I fix it, but I cant add a test for https://github.com/vuejs/vetur/issues/2300#issuecomment-696229034.

server/src/modes/script/javascript.ts
```diff
@@ -760,6 +761,13 @@ function getFormatCodeSettings(config: any): ts.FormatCodeSettings {
   };
 }
 
+function getTsTriggerCharacter(triggerChar: string) {
```
octref, line 764: Nice trick here!

octref (approved):
Great fix 👍

test/lsp/features/completion/script.test.ts
```diff
@@ -2,9 +2,10 @@ import { position } from '../../../util';
 import { testCompletion } from '../../../completionHelper';
 import { getDocUri } from '../../path';
 
-describe('Should autocomplete for <script>', () => {
+describe.only('Should autocomplete for <script>', () => {
```
octref, line 5: Not sure why you are adding this here though

yoyo930021: Oh, forget to remove debug code. XD

server/src/modes/script/javascript.ts
```diff
@@ -760,6 +761,15 @@ function getFormatCodeSettings(config: any): ts.FormatCodeSettings {
...
+  const legalChars = ['@', '#', '.', '"', "'", '`', '/', '<'] as const;
+  type LegalChars = typeof legalChars[number];
+  if (legalChars.includes(triggerChar as LegalChars)) {
+    return triggerChar as LegalChars;
+  }
+  return undefined;
+}
```
rchl, line 771: You are using `as` a lot :)

Wouldn't it be better to do this:

```ts
function getTsTriggerCharacter(triggerChar: ts.CompletionsTriggerCharacter) {
  const legalChars: ts.CompletionsTriggerCharacter[] = ['@', '#', '.', '"', "'", '`', '/', '<'];
  if (legalChars.includes(triggerChar)) {
    return triggerChar;
  }
  return undefined;
}
```

and then just use `as ts.CompletionsTriggerCharacter` for the parameter at call site?

yoyo930021: If you want to care about this,

```typescript
// Parameter must to be string, Otherwise I don't like it semantically.
function getTsTriggerCharacter(triggerChar: string) {
  const legalChars = ['@', '#', '.', '"', "'", '`', '/', '<'];
  if (legalChars.includes(triggerChar)) {
    return triggerChar as ts.CompletionsTriggerCharacter;
  }
  return undefined;
}
```

octref: @yoyo930021 I'll take this change and also remove the extra `.only`.

PR #2357, opened 2020-10-07
Support optional deprecated when completion

### LINKED ISSUE #2326: Bug with optional chaining autocompletion with strictNullCheck enabled

## Info

- Platform: macOS
- Vetur version: 0.28.0
- VS Code version: 1.49

## Problem

Normally since TS 3.7 with `strictNullCheck: true`, VSCode would behave like this when autocompleting a property of a maybe undefined object (in a TS file), and activated by the setting `typescript.suggest.includeAutomaticOptionalChainCompletions`

Like here

However in a `.vue` file, this behaviour is happening

`post` is of type `Post | undefined`, so it would normally add the `?.`, but it even removes the `.`

Bug vetur

yoyo930021:
Fixed #2326

### DIFF
--- a/server/src/modes/script/javascript.ts
+++ b/server/src/modes/script/javascript.ts
@@ -25,7 +25,8 @@ import {
   CodeAction,
   CodeActionKind,
   WorkspaceEdit,
-  FoldingRangeKind
+  FoldingRangeKind,
+  CompletionItemTag
 } from 'vscode-languageserver-types';
 import { LanguageMode } from '../../embeddedSupport/languageModes';
 import { VueDocumentRegions, LanguageRange } from '../../embeddedSupport/embeddedSupport';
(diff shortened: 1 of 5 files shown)


server/src/modes/script/javascript.ts
```diff
@@ -163,17 +164,19 @@ export async function getJavascriptMode(
...
-            label,
+            label: label + (kinds.optional ? '?' : ''),
+            tags: kinds.deprecated ? [CompletionItemTag.Deprecated] : undefined,
             detail,
             filterText: getFilterText(entry.insertText),
             sortText: entry.sortText + index,
-            kind: toCompletionItemKind(entry.kind),
-            textEdit: range && TextEdit.replace(range, entry.name),
+            kind: kinds.color ? CompletionItemKind.Color : toCompletionItemKind(entry.kind),
+            textEdit: range && TextEdit.replace(range, entry.insertText || entry.name),
```
octref, line 179: Let's mostly follow VS Code's source code's logic, so in the future it's easy to update.
Besides, there's too many condition in a single statement so we should simplify it.

octref (approved):

PR #2378, opened 2020-10-14
Vetur config file RFC

yoyo930021:
Vetur configuration file.

[Rendered](https://github.com/vuejs/vetur/blob/vetur-config-file-rfc/rfcs/001-vetur-config-file.md)

### DIFF
--- a/rfcs/001-vetur-config-file.md
+++ b/rfcs/001-vetur-config-file.md
@@ -36,7 +36,7 @@ module.exports = {
       // Where is TypeScript config file in the project?
       // It is relative to root property.
       tsconfig: './tsconfig.json',
-      // **required** default: `[]`
+      // **optional** default: `[]`
       // Register globally Vue component glob.
       // If you set it, you can get completion by that components.
       // It is relative to root property.
--- a/rfcs/001-vetur-config-file.md
+++ b/rfcs/001-vetur-config-file.md
@@ -1,0 +1,212 @@
+- Start Date: 2020-10-14
+- Target Major Version: 2.x & 3.x
+- Reference Issues: https://github.com/vuejs/vetur/issues/2325, https://github.com/vuejs/vetur/issues/2243, https://github.com/vuejs/vetur/issues/1635, https://github.com/vuejs/vetur/issues/815, https://github.com/vuejs/vetur/issues/424
+- Implementation PR: (leave this empty)
+
+# Summery
+A new configuration file for Vetur and VTI
+
+# Basic example
+
+```javascript
+// vetur.config.js
+/** @type {import('vti').VeturConfig} */
+module.exports = {
+  // **optional** default: `{}`
+  // override vscode settings part
+  // Notice: It only affects the settings used by Vetur.
+  settings: {
+    "vetur.useWorkspaceDependencies": true,
+    "vetur.experimental.templateInterpolationService": true
+  },
+  // **optional** default: `[{ root: './' }]`
+  // support monorepos
+  repos: [
+    {
+      // **required**
+      // Where is your project?
+      // It is relative to `vetur.config.js`.
+      root: './packages/repo1',
+      // **optional** default: `'package.json'`
+      // Where is `package.json` in the project?
+      // We use it to determine the version of vue.
+      // It is relative to root property.
+      package: './package.json',
+      // **optional** default: `'package.json'`
+      // Where is TypeScript config file in the project?
+      // It is relative to root property.
+      tsconfig: './tsconfig.json',
+      // **required** default: `[]`
+      // Register globally Vue component glob.
+      // If you set it, you can get completion by that components.
+      // It is relative to root property.
+      // Notice: It won't actually do it. You need to use `require.context` or `Vue.component`
+      globallyComponents: [
+        './src/components/**/*.vue'
+      ]
+    }
+  ]
+}
+```
+
+# Motivation
+- The VTI need some method to set configuration.
+- The monorepo need a baseline or logic.
+- VSCode workspace config is own by vscode. I don't want to commit it in git.
+- Shared settings required for team projects.
+- Register global vue component.
+
+# Detailed design
+
+## Noun
+- Vetur: a VSCode extension for Vue support.
+- VTI: a CLI for Vue file type-check, diagnostics or some feature.
+- VLS: vue language server, The core of everything. It is base on [language server protocol](https://microsoft.github.io/language-server-protocol/).
+
+## Config file spec
+- All path formats are used with `/`.
+  > Helpful for cross-platform use project.
+- Only support commonjs format.
+  > We can use it quickly and directly.
+- Use pure JavaScript.
+  > Same as above.
+  > You can get typings like `@JSDoc`.
+- UTF-8 charset
+
+## How to use
+
+### VTI
+You can use it to override VTI default settings.
+```bash
+vti `action`
+vti -c vetur.config.js `action`
+vti --config vetur.config.js `action`
+```
+
+### Vetur
+This profile takes precedence over vscode setting.
+It will find it when Vetur initialization.
+If it isn't exist, It will use `{ settings: {}, monorepos: [{ root: './' }] }`.
+This will ensure consistency with past behavior.
+
+PS. Each root can have its own vetur.config.js in VSCode Multi root feature.
+
+## Property detail
+
+### Definition
+```typescript
+type Glob = string
+
+export interface VeturConfig {
+  settings?: { [key: string]: boolean | string | Enum },
+  monorepos?: Array<{
+    root: string,
+    package?: string,
+    tsconfig?: string,
+    globallyComponents?: Glob[]
+  }>
+}
+```
+
+### `settings`
+Incoming to vue language server config.
+
+In VLS, it will merge (vscode setting or VTL default config) and vetur.config.js `settings`.
+```typescript
+import _ from 'lodash'
+
+// original vscode config or VTI default config
+const config: VLSFullConfig = params.initializationOptions?.config
+  ? _.merge(getDefaultVLSConfig(), params.initializationOptions.config)
+  : getDefaultVLSConfig();
+
+// From vetur.config.js
+const veturConfig = getVeturConfigInWorkspace()
+// Merge vetur.config.js
+Object.keys(veturConfig.setting).forEach((key) => {
+  _.set(config, key, veturConfig.setting[key])
+})
+```
+
+Notice: It only affects the settings used by Vetur.
+For example, we use `typescript.preferences.quoteStyle` in Vetur. so you can set it.
+But it don't affect original TypeScript support in VSCode.
+
+### `repos`
+The monorepo need a baseline or logic.
+Possible options are `package.json` or `tsconfig.js`.
+But both are used for node and typescript projects.
+We're likely to waste unnecessary resources on things we don't need.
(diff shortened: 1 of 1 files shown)


victorgarciaesgi:
This is really great!

I have nothing to comment at the moment, everything seems ideal (I don't use monorepos thought so I can't debate on this)

IlCallo:
> Put configuration to `vue.config.js`

I'd exclude this alternative, as there are other CLIs (Nuxt, Quasar, etc) which needs Vetur

Into "properties details > definition" you use `monorepos` property name instead of `repos`

Have you checked other frameworks config files for tasks similar to this? Angular/React/whatever? If yes, is it possible to have link to relevant resources/inspiration?

yoyo930021:
> > Put configuration to `vue.config.js`
> 
> I'd exclude this alternative, as there are other CLIs (Nuxt, Quasar, etc) which needs Vetur
> 
> Into "properties details > definition" you use `monorepos` property name instead of `repos`
> 
> Have you checked other frameworks config files for tasks similar to this? Angular/React/whatever? If yes, is it possible to have link to relevant resources/inspiration?

Thanks.
I fixed wrong text in RFC.

My first inspiration came from `svelte.config.js`.
https://github.com/sveltejs/svelte/issues/1101 https://github.com/sveltejs/language-tools/blob/3139ccd5b4306a914be4ba64b0c036fcd8888f78/packages/svelte-vscode/README.md#generic-setup
The Svelte VScode extension read this file.
But they don't really have the support of monorepo.

The React and Angular are all officially supported by TypeScript for monorepo support.
They are all based on TypeScript program `tsconfig.json`,
And they don't have their own unique files to deal with.
PS. TypeScript is direct support for `jsx`.

yoyo930021:
I extend `repos[].globallyComponents` in RFC.
More consideration for customizable component names.

andrewisaburden:
Grammar for `globallyComponents` is not correct; perhaps `globalComponents` or `globallyRegisteredComponents`?

yoyo930021:
> Grammar for `globallyComponents` is not correct; perhaps `globalComponents` or `globallyRegisteredComponents`?

I rename to `globalComponents`.

rfcs/001-vetur-config-file.md
```diff
@@ -0,0 +1,225 @@
...
+
+## Property detail
+
+### Definition
+```typescript
+type Glob = string
+
+export interface VeturConfig {
+  settings?: { [key: string]: boolean | string | Enum },
+  repos?: Array<{
```
znck, line 106: `repos` is more connected to git repositories, and I get it, it's coming from monorepos but I feel it's not conveying the purpose.

Suggestion: We can call it `projects`.  (typescript has project references and that makes it some what similar)

rfcs/001-vetur-config-file.md
```diff
@@ -0,0 +1,225 @@
...
+    root: string,
+    package?: string,
+    tsconfig?: string,
+    globalComponents?: Array<Glob | { name: string, path: string }>
```
znck, line 110: How about `"extends"` option to share configurations?

e.g.
```js
module.export = {
  projects: [{
    extends: ["vuetify"],
    globalComponents: ["./src/components/**/*.vue"]
  }]
}
```

rfcs/001-vetur-config-file.md
```diff
@@ -0,0 +1,225 @@
...
+- Vetur: a VSCode extension for Vue support.
+- VTI: a CLI for Vue file type-check, diagnostics or some feature.
+- VLS: vue language server, The core of everything. It is base on [language server protocol](https://microsoft.github.io/language-server-protocol/).
+
+## Config file spec
+- All path formats are used with `/`.
+  > Helpful for cross-platform use project.
+- Only support commonjs format.
+  > We can use it quickly and directly.
+- Use pure JavaScript.
```
znck, line 71: What is the reason for choosing `.js` over `.json`?

antfu:
`globalComponents` will help a lot to me! Really looking forward to this RFC get implemented! 👍

rfcs/001-vetur-config-file.md
antfu: Maybe workspaces?

yoyo930021: I agree. `projects` is better.

rfcs/001-vetur-config-file.md
yoyo930021: I don't understand what this setting means.
I think if we're going to share configurations, we can only use JS. Like `require`.

PS. When you use `vuetify` on package.json, we use component date to completion.
So you don't really need to define `vuetify` here!

rfcs/001-vetur-config-file.md
yoyo930021: Native JSON does not support commenting.
If you want to share your settings, you can use the JS mechanism directly.
Even if you want to support plugins in the distant future, this format will be easier to handle.

rfcs/001-vetur-config-file.md
sqal: I like the "workspaces" name or maybe workingDirectories (stolen from eslint extension 😜)

rfcs/001-vetur-config-file.md
znck: Similar to tsconfig but it doesn't makes sense in cjs.

rfcs/001-vetur-config-file.md
znck: workspaces also sounds good as package managers have those.

yoyo930021: The `workspaces` word is also used for multi roots in VSCode.
It's a bit confusing.

octref: As long as it's clear what it meant, it's ok, so +1 for projects.

rfcs/001-vetur-config-file.md
octref: @znck I don't feel we need this. When you have:
- `vuetify` listed in pointed package.json's dependencies
- `node_modules/vuetify/package.json` has `vetur` key

Everything should work from there.

Unless you meant something else.

znck: We need to support opt-in for external global components. (user can register, only a few components)

rfcs/001-vetur-config-file.md
```diff
@@ -0,0 +1,225 @@
...
+      // **optional**
+      // Where is TypeScript config file in the project?
+      // It is relative to root property.
+      tsconfig: './tsconfig.json',
+      // **required** default: `[]`
+      // Register globally Vue component glob.
+      // If you set it, you can get completion by that components.
+      // It is relative to root property.
+      // Notice: It won't actually do it. You need to use `require.context` or `Vue.component`
+      globalComponents: [
```
octref, line 44: Probably optional here? No value = `[]`.

rfcs/001-vetur-config-file.md
octref: Probably also allow a shorthand form, so

```ts
Array<string | {
    root: string,
    package?: string,
    tsconfig?: string,
    globalComponents?: Array<Glob | { name: string, path: string }
>
```

Where `string` value is `root`.

rfcs/001-vetur-config-file.md
octref: @znck What's a use case for that? If I understand you correctly, do you mean for the case when you have 2 UI lib and only want auto-completion for one of them (or only a few components out of a full library)? I feel there isn't much downside to just registering a whole library.

octref (changes requested):
LGTM, except 2 small things:

- `repos` -> `projects`
- Allow `projects` to be an array of either the shortform (just a string pointing to `root`) or the longer form (`root/package/tsconfig/globalComponents`).

rfcs/001-vetur-config-file.md
znck: Mainly for type checking global components.

rfcs/001-vetur-config-file.md
yoyo930021: It's a good idea.
I changed the RFC.

rfcs/001-vetur-config-file.md
yoyo930021: It's a error.
I fix it.

rfcs/001-vetur-config-file.md
yoyo930021: I would prefer to use cjs directly.

module.export = {
  projects: [{
    globalComponents: [
      ...require('vuetify').components,
      "./src/components/**/*.vue"
    ]
  }]
}

yoyo930021: I also think there's no point in adding an external lib.
We don't know the details of external libs.
It may be difficult to get the correct TypeScript type when no correct `tsconfig.json`.

octref (approved):
LGTM, thanks for addressing the feedback.

rfcs/001-vetur-config-file.md
znck: In the TS plugin, I am using [this](https://github.com/znck/vue-developer-experience/blob/fb9655dfcb4b24a015e1dae76e1df14b241b6d0b/packages/analyze/src/component.ts#L83-L89) to infer component meta. (It has a different structure compared to vetur's spec but includes verbose static information).

znck:
How would named component from external module work?

```js
important { Foo } from 'foo'

app.component('MyFoo', Foo)
```

yoyo930021:
> How would named component from external module work?
> 
> ```js
> important { Foo } from 'foo'
> 
> app.component('MyFoo', Foo)
> ```

This RFC don't solves the problem.
Vetur currently uses component data to do this.
In the long run, we need a type-based mechanism.
I think we can discuss a new RFC about this.

znck:
In the TS plugin, I am using the following data structure for the project config.

```ts
export interface ImportSource {
  moduleName: string
  exportName?: string
}

export interface ProjectConfig {
  root: string
  package?: string
  tsconfig?: string
  globalComponents: Array<string | Record<string, string | ImportSource>>
}
```

znck:
What is the purpose of `tsconfig` option?

yoyo930021:
> In the TS plugin, I am using the following data structure for the project config.
> 
> ```ts
> export interface ImportSource {
>   moduleName: string
>   exportName?: string
> }
> 
> export interface ProjectConfig {
>   root: string
>   package?: string
>   tsconfig?: string
>   globalComponents: Array<string | Record<string, string | ImportSource>>
> }
> ```

I think we can keep it that way for now.
If vetur introduces such an approach, it will have to face the problem of historical component data.

yoyo930021:
> What is the purpose of `tsconfig` option?

The vetur will use this option file to open the ts language service.

znck:
TS Plugin can only use `globalComponents`, everything else is inherited from the TypeScript project.

yoyo930021:
> TS Plugin can only use `globalComponents`, everything else is inherited from the TypeScript project.

You may also want to refer to the `package`.
It can be used to know the Vue version of the project.
What external libraries are used?

Lanseria:
Can you set the vetur version here, so that accidental upgrades don't cause problems?

yoyo930021:
> Can you set the vetur version here, so that accidental upgrades don't cause problems?

It's practically impossible.
You only can use `vetur.dev.vlsPath` option to pin vls version.

PR #2379, opened 2020-10-14
changed warming to warning

travisboss:
Changed this line: Vetur will show a **warming** for `<simple>` and an error for `<complex>`.

To this line: Vetur will show a **warning** for `<simple>` and an error for `<complex>`.


octref (approved):
Guess I read too much about climate change 😅

PR #2389, opened 2020-10-16
Refactor to setupConfigure for #2388

### LINKED ISSUE #2388: Vetur auto formatting not working on .vue files until disabled and then re-enabled in preferences
- [x ] I have searched through existing issues
- [x ] I have read through [docs](https://vuejs.github.io/vetur)
- [x ] I have read [FAQ](https://github.com/vuejs/vetur/blob/master/docs/FAQ.md)

## Info

- Platform: macOS Catalina
- Vetur version: 0.28.0
- VS Code version: 1.50.1

## Problem
After launching VSCode, Vetur does not format .vue files until going into vscode preferences > extensions > vetur > toggle `Enable/Disable the Vetur document formatter` off and then on again.  After toggling, it works as expected.
Output from panel, both before and after toggling the mentioned setting so it appears Vetur is loaded when VSCode launches:
```
Loaded bundled typescript@4.0.2.
Vetur initialized
```

## Reproducible Case

Unsure how I can reproduce this, but there is at least one other user on the Discord vueland eslint channel who has the same/similar problem.

yoyo930021:
Fixed #2388 

It also conveniently fixes a problem where incorrect settings could be imported.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - 🙌 Slim syntax highlighting. Thanks to contribution from [@Antti](https://github.com/Antti).
 - 🙌 Stop computing outdated diagnostics with CancellationToken. Thanks to contribution from [@yoyo930021](https://github.com/yoyo930021). #1263 and #2332.
 - 🙌 Fix error when optional camel-cased props are missing. Thanks to contribution from [@yoyo930021](https://github.com/yoyo930021). #2314 and #2342.
+- 🙌 Fix Vetur formatting not working. Thanks to contribution from [@yoyo930021](https://github.com/yoyo930021). #2388 and #2389.
 
 ### 0.28.0 | 2020-09-23 | [VSIX](https://marketplace.visualstudio.com/_apis/public/gallery/publishers/octref/vsextensions/vetur/0.28.0/vspackage)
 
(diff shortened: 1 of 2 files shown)


octref (approved):
LGTM 👍

rchl:
I'd like to better understand the changes done here to fix it properly in #2436.

Why was it not enough to run `setupDynamicFormatters()` from `onDidChangeConfiguration` notification?

Sending it when handling `initialize` request is too early.

We could send it from `initialized` but I'm not sure why from `onDidChangeConfiguration()` was not enough since the client is pretty much guaranteed to send it early after initialize.

Was the problem more related to passing a partial config to `setupDynamicFormatters`?

yoyo930021:
> I'd like to better understand the changes done here to fix it properly in #2436.
> 
> Why was it not enough to run `setupDynamicFormatters()` from `onDidChangeConfiguration` notification?
> 
> Sending it when handling the `initialize` request, as done here, is too early.
> 
> We could send it from `initialized` but I'm not sure why from `onDidChangeConfiguration()` was not enough since the client is pretty much guaranteed to send it early after initialize.
> 
> Was the problem more related to passing a partial config to `setupDynamicFormatters`?

I think it's a try.
My first suspicion was that it wouldn't be triggered `onDidChangeConfiguration` on startup.
I guess we could rollup back this part.

rchl:
OK, good. That means I can safely revert that logic. Thanks.

PR #2403, opened 2020-10-23
Fix crash when wrong tsconfig

yoyo930021:
Fixing an incorrect tsconfig can cause a crash.

Mark `vue-editor-bridge` as external so that it does not spew errors that are not in the project.


server/src/services/typescriptService/serviceHost.ts
```diff
@@ -271,7 +271,8 @@ export function getServiceHost(
           if (name === bridge.moduleName) {
             return {
               resolvedFileName: bridge.fileName,
-              extension: tsModule.Extension.Ts
+              extension: tsModule.Extension.Ts,
+              isExternalLibraryImport: true
```
octref, line 275: Don't know where you found this. Next time would be great if you can give a bit more explanation or point to the code. Thanks!

octref (approved):

server/src/services/typescriptService/serviceHost.ts
yoyo930021: I only think why no include file error in `node_modules`.
So there must be a way to mark this.
And I found this.
https://github.com/microsoft/TypeScript/blob/bcbe1d763823eacd4b252c904e77761a6b8636a1/src/compiler/types.ts#L6216

PR #2404, opened 2020-10-27
Add slim syntax highlighting

Antti:
Add [Slim](http://slim-lang.com/) syntax highlighting via [Slim extension](https://marketplace.visualstudio.com/items?itemName=sianglim.slim)

### DIFF
--- a/syntaxes/vue-generated.json
+++ b/syntaxes/vue-generated.json
@@ -7,6 +7,49 @@
   "uuid": "5512c10d-4cc5-434c-b8fc-53b912f55ab3",
   "patterns": [
     {
+      "begin": "(<)(slim)",
+      "beginCaptures": {
+        "1": {
+          "name": "punctuation.definition.tag.begin.html"
+        },
+        "2": {
+          "name": "entity.name.tag.style.html"
+        }
+      },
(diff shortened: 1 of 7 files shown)


reneklacan (approved):
looks great :heart:

majtymit:
lovely 👍

octref (approved):
Good work, nothing need to change. Thank you 👍

syntaxes/vue-generated.json
```diff
@@ -6,6 +6,49 @@
   ],
   "uuid": "5512c10d-4cc5-434c-b8fc-53b912f55ab3",
   "patterns": [
+    {
```
yoyo930021, line 9: @Antti
I notice this block code.
it isn't generated from yaml.

Is this code necessary?

Antti: @yoyo930021  I have only edited the YAML file and run the generator, so I believe it was generated from it. At least at that point of time.

yoyo930021: That's weird.
Now when running will disappear.
I don't think it should be necessary and I will remove it.

PR #2426, opened 2020-11-02
Fix optional chaining in template

### LINKED ISSUE #2423: Optional Chaning does not work on template

## Info

- Platform: Windows 10
- Vetur version: 0.28.0
- VS Code version: 1.50.1

## Problem
I'm using Vue 3 and when I use optional chaning in template it shows a error saying "object is possibly 'null' or 'udefined' ". I think it should not show the error because when I use null coalescing operator, no error is shown.

**Error Message**:  "Object is possibly 'null' or 'undefined'.Vetur(2533)"

image

yoyo930021:
Fixed #2423


octref (approved):
LGTM, just added a unit test. I'd prefer to add unit test whenever possible since they run much faster.

Issue #2390, opened 2020-10-17
Nativescript-vue v-template support


farfromrefug:
When using Vue in Nativescript we have a custom v component called `v-template` https://nativescript-vue.org/en/docs/utilities/v-template/
It kind of work like v-for. Kind of! the parent of the template defines a `:items` property which is an array like object.
Then inside the template you can use `item` which is an object of `items`
Do you think you could add support for it so that we get completion on template?

Thanks

yoyo930021:
Vetur support custom component date for custom framework or libs.

You need to add this:
1. Add component data files. Example: https://github.com/vuejs/vue-router/tree/dev/vetur
2. Add key in `package.json`. Example: https://github.com/vuejs/vue-router/blob/dev/package.json#L24
Doc: https://vuejs.github.io/vetur/component-data.html#workspace-component-data

yoyo930021:
The tricky thing here is the `if` prop question.
If you want to get it right type check in template interpolation, We need to requiring special handling of `if` prop.

farfromrefug:
@yoyo930021 thanks ! tried that and got `v-template` to be "recognized`. However what i would like is for vetur to "know" that item is an item of the array defined in `:items` above the template, so that i get type and completion on the `item` object. Just like it happens with `v-for`. Do you think this is possible?

yoyo930021:
> @yoyo930021 thanks ! tried that and got `v-template` to be "recognized`. However what i would like is for vetur to "know" that item is an item of the array defined in `:items`above the template, so that i get type and completion on the`item`object. Just like it happens with`v-for`. Do you think this is possible?

This question is the same as if.
I think this is not easy to handle.
It is difficult for us to deal with this particular component.
Call @ktsn for check.

Why isn’t this API designed as a slot?
This way you can avoid this problem perfectly.

farfromrefug:
@rigor789 do you a reason not to use slots as @yoyo930021 suggests?

yoyo930021:
PS. scoped slot type check is not support now.
But in the future we will support.

PR #2429, opened 2020-11-02
Update FAQ.md

baybal:


octref:
You should follow #2390. I don't think it's worth it to add a FAQ for this issue.

baybal:
@octref this a completely separate issue.

What I was talking about is when intellisense is completely inoperable because Vue package is not installed by default.

octref:
Shouldn't you open a PR for the nativescript-vue template you are using?

PR #2431, opened 2020-11-03
Fix/2430 sync modifier prop type checking

### LINKED ISSUE #2430: v-bind with .sync modifier and Vue template interpolation/type checking gives TypeScript error

## Problem

If `vetur.experimental.templateInterpolationService: true`, and using a v-bind with .sync modifier, TypeScript reports: 

`Property 'item' is missing in type '{ "item.sync": string; }' but required in type '{ item: string;  }'`

## Reproducible Case

```
<template>
    <SomeComponent :item.sync="item" />
</template>
```

SomeComponent.vue:

```
<script lang="ts">
import Vue from 'vue';

export default Vue.extend({
  props: {
    item: String,
  }
});
</script>
```

I have a fix, and will make a PR shortly.

It is caused by v-bind and v-on modifiers being appended in the same manner (originally to prevent duplicates causing errors from v-on chained modifiers). v-bind modifiers should always be culled from the node name, because we cannot bind one Vue prop twice (duplicates for v-bind are an error), and so that we aren't trying to find a prop with name "item.sync" rather than "item".

andrewisaburden:
Fixes #2430


server/src/services/typescriptService/transformTemplate.ts
```diff
@@ -251,7 +251,7 @@ export function getTemplateTransformFunctions(ts: T_TypeScript, childComponentNa
         // Attribute name is specified
         // e.g. v-bind:value="foo"
         const fullName =
-          dir.key.modifiers.length === 0
+          dir.key.modifiers.length === 0 || !isVOn(dir)
```
yoyo930021, line 254: I think `&& isVBind(dir)` more intuitive.

yoyo930021 (changes requested):

server/src/services/typescriptService/transformTemplate.ts
andrewisaburden: @yoyo930021 I agree, and initially that's what I did. However, for reasons that are not clear to me, using `isVBind` with the typeguard of `: node is AST.VDirective` results in line 256 throwing this TypeScript error: `Property 'key' does not exist on type 'never'.` for the section `...dir.key.modifiers.map`.

This doesn't make sense to me, because `dir` is already of type `AST.VDirective` in the function parameters. I couldn't figure out a solution, if you know one please let me know.

andrewisaburden: After reading a little more, typeguards that are type predicates *only* do one thing: confirm the type is guarded.

`isVBind` actually should be split into two methods, one for `isVDirective` and one for `isVBind`. `isVDirective` should be a type guard, e.g. `function isVDirective(node: AST.VAttribute | AST.VDirective): node is AST.VDirective {`
and `isVBind` should not be a type guard: `function isVBind(node: AST.VAttribute | AST.VDirective): boolean {`

yoyo930021 (approved):

yoyo930021:
@ktsn If you have time, you can take a look.
I think I'll merge first. If you have any questions, let me know.

PR #2458, opened 2020-11-11
Fix vls initialize is slow

### LINKED ISSUE #2453: VLS initialize is slow
Ref: https://github.com/vuejs/vetur/pull/2301#issuecomment-724723344

yoyo930021:
Fixed #2453

### DIFF
--- a/server/src/services/dependencyService.ts
+++ b/server/src/services/dependencyService.ts
@@ -14,30 +14,49 @@ import { performance } from 'perf_hooks';
 import { logger } from '../log';
 
 const readFileAsync = util.promisify(fs.readFile);
+const accessFileAsync = util.promisify(fs.access);
 
-async function findAllPackages(workspacePath: string, moduleName: string) {
+async function createNodeModulesPaths(workspacePath: string) {
   const startTime = performance.now();
-  const packages = await fg(`**/node_modules/${moduleName}/package.json`, {
-    cwd: workspacePath,
-    absolute: true,
-    unique: true
-  }).then(filePaths =>
-    Promise.all(
-      filePaths.map(filePath =>
-        readFileAsync(filePath, { encoding: 'utf8' }).then(content => {
-          const info = JSON.parse(content) as { name: string; version: string; main: string };
+  const nodeModules = (
+    await fg('**/node_modules', {
+      cwd: workspacePath,
+      absolute: true,
+      unique: true,
+      onlyDirectories: true,
+      onlyFiles: false
+    })
+  ).filter(el => el.match(/node_modules/g)?.length === 1);
 
-          return {
-            name: info.name,
-            dir: path.dirname(filePath),
-            version: info.version,
-            module: require(path.resolve(path.dirname(filePath), info.main))
-          };
-        })
-      )
-    )
-  );
-  logger.logInfo(`Try to find ${moduleName} in ${workspacePath}. - ${Math.round(performance.now() - startTime)}ms`);
+  logger.logInfo(`Find node_modules paths in ${workspacePath} - ${Math.round(performance.now() - startTime)}ms`);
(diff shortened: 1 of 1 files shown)


server/src/services/dependencyService.ts
```diff
@@ -14,30 +14,50 @@ import { performance } from 'perf_hooks';
 import { logger } from '../log';
 
 const readFileAsync = util.promisify(fs.readFile);
+const existsFileAsync = util.promisify(fs.existsSync);
```
rchl, line 17: Doesn't make sense to promisify sync version IMO. You'd probably want to promisy `fs.exists` instead (or `fs.stat` rather since `fs.exists` is deprecated).

yoyo930021: Wow, thanks a lot.
I overtyped `Sync`.

server/src/services/dependencyService.ts
```diff
@@ -14,30 +14,50 @@ import { performance } from 'perf_hooks';
 import { logger } from '../log';
 
 const readFileAsync = util.promisify(fs.readFile);
+const existsFileAsync = util.promisify(fs.exists);
```
rchl, line 17: Would still recommend using non-deprecated API.

rchl:
I'm seeing multiple instances of `[INFO ] Loaded typescript@4.0.5 from /Users/...` during initialize.

It's not only loading typescript from root of the `node_modules` but also within the dependencies, if those include typescript.

I know that dependencies should generally not include typescript as in `dependencies` but I have a couple of packages linked. But in any case, linked or not, should this be optimized to only try to load from the root of `node_modules`?

yoyo930021:
> I'm seeing multiple instances of `[INFO ] Loaded typescript@4.0.5 from /Users/...` during initialize.
> 
> It's not only loading typescript from the root of the `node_modules` but also within the dependencies if those include typescript.
> 
> I know that dependencies should generally not include typescript in `dependencies` but I have a couple of packages linked. But in any case, linked or not, should this be optimized to only try to load from the root of `node_modules`?

This code was prepared for monorepo.
People should be able to use different versions of TypeScript in their projects.
Although the TypeScript ecosystem doesn't seem to allow it.

rchl:
With the monorepo setup, the intention is that each sub-project will have its own `package.json`, right? So my suggestion should still be OK.

yoyo930021:
> With the monorepo setup, the intention is that each sub-project will have its own `package.json`, right? So my suggestion should still be OK.
> 
> And it's not only about typescript. Other dependencies like prettier will also be searched and discovered in some deep sub-dependencies.

Your suggestion is good.
But this logic was prepared for the other libraries that came before it.
https://github.com/vuejs/vetur/blob/f37b9846873d613d0818d09268663618681395cb/server/src/utils/prettier/requirePkg.ts#L17
The main thing is that I want to be as consistent as possible with the require behavior.

rchl:
But `require` wouldn't find dependency `c` located at `node_modules/a/node_modules/c` unless it would be called from `node_modules/a/index.js`, for example. So I'm not sure how current behavior is equivalent to `require`.

Note that I might not be seeing the whole picture...

yoyo930021:
> But `require` wouldn't find dependency `c` located at `node_modules/a/node_modules/c` unless it would be called from `node_modules/a/index.js`, for example. So I'm not sure how current behavior is equivalent to `require`.
> 
> Note that I might not be seeing the whole picture...

You're right.

yoyo930021:
@rchl Thanks a lot.
Can you help me test your project?

rchl:
Looks good with the latest changes. It didn't find any extraneous typescript dependencies (only the single, expected one) or other dependencies. And it was relatively fast (~2s` for `initialize` response).

Issue #2469, opened 2020-11-13
support yarn pnp mode


yoyo930021:
- [x] I have searched through existing issues

## Feature Request
Although we allow custom typescript to solve this problem,
But reading deps in the project isn't working in pnp mode.

PR #2478, opened 2020-11-16
feat: add Yarn PnP support

merceyz:
**What's the problem this PR addresses?**

The tag providers can't be located under Yarn PnP due to hard coding `node_modules`

Fixes https://github.com/vuejs/vetur/issues/2469

**How did you fix it?**

- Use require.resolve to locate tag providers
- Setup the PnP hook from the workspace if it exists

### DIFF
--- a/server/src/modes/template/tagProviders/externalTagProviders.ts
+++ b/server/src/modes/template/tagProviders/externalTagProviders.ts
@@ -52,10 +52,13 @@ export function getDependencyTagProvider(workspacePath: string, depPkgJson: any)
     return null;
   }
 
-  const tagsPath = findConfigFile(workspacePath, path.join('node_modules/', depPkgJson.name, depPkgJson.vetur.tags));
+  const tagsPath = findConfigFile(
+    workspacePath,
+    require.resolve(path.join(depPkgJson.name, depPkgJson.vetur.tags), { paths: [workspacePath] })
+  );
   const attrsPath = findConfigFile(
     workspacePath,
-    path.join('node_modules/', depPkgJson.name, depPkgJson.vetur.attributes)
+    require.resolve(path.join(depPkgJson.name, depPkgJson.vetur.attributes), { paths: [workspacePath] })
   );
 
   try {
--- a/server/src/modes/template/tagProviders/externalTagProviders.ts
+++ b/server/src/modes/template/tagProviders/externalTagProviders.ts
@@ -52,18 +52,15 @@ export function getDependencyTagProvider(workspacePath: string, depPkgJson: any)
     return null;
   }
 
-  const tagsPath = require.resolve(path.join(depPkgJson.name, depPkgJson.vetur.tags), { paths: [workspacePath] });
-  const attrsPath = require.resolve(path.join(depPkgJson.name, depPkgJson.vetur.attributes), {
-    paths: [workspacePath]
-  });
-
   try {
-    if (tagsPath && attrsPath) {
-      const tagsJson = JSON.parse(fs.readFileSync(tagsPath, 'utf-8'));
-      const attrsJson = JSON.parse(fs.readFileSync(attrsPath, 'utf-8'));
-      return getExternalTagProvider(depPkgJson.name, tagsJson, attrsJson);
-    }
-    return null;
+    const tagsPath = require.resolve(path.join(depPkgJson.name, depPkgJson.vetur.tags), { paths: [workspacePath] });
+    const attrsPath = require.resolve(path.join(depPkgJson.name, depPkgJson.vetur.attributes), {
+      paths: [workspacePath]
+    });
+
+    const tagsJson = JSON.parse(fs.readFileSync(tagsPath, 'utf-8'));
+    const attrsJson = JSON.parse(fs.readFileSync(attrsPath, 'utf-8'));
+    return getExternalTagProvider(depPkgJson.name, tagsJson, attrsJson);
   } catch (err) {
     return null;
   }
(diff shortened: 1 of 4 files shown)


merceyz:
Would it be fine if I changed https://github.com/vuejs/vetur/commit/e88c26b3f5e02ad42c2d49b2eb06eafdb86218dd to use `require.resolve` instead?

yoyo930021:
> Would it be fine if I changed
> 
> https://github.com/vuejs/vetur/blob/03a4373ffb7cae8e07d313a886b1a72a386799f3/server%2Fsrc%2Fservices%2FdependencyService.ts#L165
> to use `require.resolve` instead?
> I guess that is supposed to half support workspaces, but at that point it should just run `require.resolve` in each workspace

I want to establish a dependency relationship in advance.
But I thought maybe in pnp it was redundant?

merceyz:
I'll omit that part for now, this makes the tag providers work under PnP

server/src/modes/template/tagProviders/externalTagProviders.ts
```diff
@@ -52,10 +52,13 @@ export function getDependencyTagProvider(workspacePath: string, depPkgJson: any)
     return null;
   }
 
-  const tagsPath = findConfigFile(workspacePath, path.join('node_modules/', depPkgJson.name, depPkgJson.vetur.tags));
+  const tagsPath = findConfigFile(
```
yoyo930021, line 55: Do we need this function when we use `require.resolve` function?

server/src/services/typescriptService/vueVersion.ts
```diff
@@ -31,7 +31,10 @@ export function inferVueVersion(workspacePath: string): VueVersion {
       return floatVersionToEnum(sloppyVersion);
     }
 
-    const nodeModulesVuePackagePath = findConfigFile(path.resolve(workspacePath, 'node_modules/vue'), 'package.json');
+    const nodeModulesVuePackagePath = findConfigFile(
+      workspacePath,
+      require.resolve('vue/package.json', { paths: [workspacePath] })
+    );
     const nodeModulesVuePackageJSON =
       nodeModulesVuePackagePath && JSON.parse(readFileSync(nodeModulesVuePackagePath, { encoding: 'utf-8' })!);
```
yoyo930021, line 39: Do we need to replace `readFileSync` to `require`?

yoyo930021 (changes requested):

server/src/services/typescriptService/vueVersion.ts
merceyz: We don't have to since the full that is resolved right above, probably shouldn't as well since `require` will cache the result so any changes wont be noticed without restarting VSCode

server/src/modes/template/tagProviders/externalTagProviders.ts
merceyz: I don't think it's needed here, no

yoyo930021: We can remove `findConfigFile`.

server/src/services/typescriptService/vueVersion.ts
yoyo930021: What is `require.resolve` result when in pnp mode?
Can it be read directly?

merceyz: It's an absolute filepath that just so happens to be within a zip file, the PnP hook takes care of adding support for reading from them directly.

Example:
```
require.resolve('lodash');
-> 
/project/.yarn/cache/lodash-npm-4.17.20-c0db62021c-7.zip/node_modules/lodash/lodash.js
```

The only PnP specific thing is to setup the `.pnp.js` file, the rest is as normal

yoyo930021: Can `fs.readFileSync` function read file in a zip file?

merceyz: Yes, as mentioned, the pnp hook adds support for it

> [...] within a zip file, the PnP hook takes care of adding support for reading from them

server/src/services/vls.ts
```diff
@@ -106,6 +107,15 @@ export class VLS {
 
     this.workspacePath = normalizeFileNameToFsPath(workspacePath);
 
+    // Enable PnP API
+    if (this.workspacePath && !process.versions.pnp && config.vetur.useWorkspaceDependencies) {
```
yoyo930021, line 111: @merceyz The latest question is `.pnp.js` position.
We try to support monorepo in this [RFC](https://github.com/vuejs/vetur/blob/vetur-config-file-rfc/rfcs/001-vetur-config-file.md#projectspackage)
Where is `.pnp.js` file? Same as `package.json`?

merceyz: It's always at the project root

yoyo930021: If I have a monorepo project,
- /
- package/project1
- package/project2
- package/project1

When I open `/`, I can get full support.
When I only open `package/project1`, It is broken.

merceyz: Correct, but that's usually how that works, it's the same for TypeScript for example

yoyo930021 (approved):
LGTM, I will merge when #2495 merged.

server/src/services/vls.ts
```diff
@@ -106,6 +107,15 @@ export class VLS {
 
     this.workspacePath = normalizeFileNameToFsPath(workspacePath);
 
+    // Enable Yarn PnP support https://yarnpkg.com/features/pnp
+    if (!process.versions.pnp) {
```
yoyo930021, line 111: What time is this variable set?

merceyz: It's set by the `.pnp.js` file when `setup` is called

PR #2558, opened 2020-12-11
Custom event auto complete

### LINKED ISSUE #2392: Template autocomplete for custom events
## Feature Request
With new features relating to custom events in Vue 3, it would be nice to have autocompletion for events based on Vue 3's custom events validation. Type validation, especially with Typescript would also be nice.

https://v3.vuejs.org/guide/component-custom-events.html#validate-emitted-events

sapphi-red:
This PR implements custom event auto complete.

Obtains data from
```ts
emits: {
  foo: () => true,
}
```
and
```ts
emits: ['foo']
```
.

image

~~Validations are not implemented.~~

close #2392

### DIFF
--- a/server/src/modes/script/componentInfo.ts
+++ b/server/src/modes/script/componentInfo.ts
@@ -221,6 +221,9 @@ function getEmits(
       if (signature) {
         const returnType = checker.getReturnTypeOfSignature(signature);
         typeString = `(${checker.typeToString(returnType)})`;
+        if (typeString === '(void)') {
+          typeString = '(undefined)';
+        }
       }
 
       if (emitInfoMap.has(name)) {
--- a/server/src/modes/script/componentInfo.ts
+++ b/server/src/modes/script/componentInfo.ts
@@ -1,8 +1,10 @@
+import _ from 'lodash';
 import type ts from 'typescript';
 import { BasicComponentInfo } from '../../config';
 import { RuntimeLibrary } from '../../services/dependencyService';
 import {
   VueFileInfo,
+  EmitInfo,
   PropInfo,
   ComputedInfo,
   DataInfo,
@@ -87,6 +89,7 @@ export function analyzeDefaultExportExpr(
   const defaultExportType = checker.getTypeAtLocation(defaultExportNode);
 
   const insertInOptionAPIPos = getInsertInOptionAPIPos(tsModule, defaultExportType, checker);
+  const emits = getEmits(tsModule, defaultExportType, checker);
   const props = getProps(tsModule, defaultExportType, checker);
   const data = getData(tsModule, defaultExportType, checker);
   const computed = getComputed(tsModule, defaultExportType, checker);
@@ -95,6 +98,7 @@ export function analyzeDefaultExportExpr(
   return {
     componentInfo: {
       insertInOptionAPIPos,
+      emits,
       props,
       data,
       computed,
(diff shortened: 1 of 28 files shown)


server/src/modes/script/componentInfo.ts
```diff
@@ -137,6 +141,92 @@ function getInsertInOptionAPIPos(
...
+     */
+    if (propertyValue.kind === tsModule.SyntaxKind.NullKeyword) {
+      return { hasValidator: false };
+    }
+
+    return { hasValidator: true };
+  }
+
+  function getClassEmits(type: ts.Type) {
+    return undefined;
```
yoyo930021, line 165: Maybe we can support this.
https://github.com/kaorun343/vue-property-decorator#Emit

yoyo930021 (changes requested):
We also miss type check in event.

Thanks for your PR.
We originally expected to change to a full type implementation.
But I think your implementation is perfect for the transition period.

sapphi-red:
I added features below!

- type check
- `@Emit` support
  - If there is multiple `@Emit` with same emit name, it infers that it retuns the union type of those return type.
  - If emits option is set also, it will not collect info from `@Emit`.

server/src/modes/script/componentInfo.ts
```diff
@@ -646,15 +828,18 @@ function getClassAndObjectInfo<C, O>(
...
+  onlyUseObjectResultIfExists = false
 ) {
   const result: Array<C | O> = [];
   if (isClassType(tsModule, defaultExportType)) {
-    result.push.apply(result, getClassResult(defaultExportType) || []);
     const decoratorArgumentType = getClassDecoratorArgumentType(tsModule, defaultExportType, checker);
     if (decoratorArgumentType) {
       result.push.apply(result, getObjectResult(decoratorArgumentType) || []);
     }
+    if (result.length === 0 || !onlyUseObjectResultIfExists) {
```
yoyo930021, line 840: We can't `result.length === 0` condition.
Class component allow this.

yoyo930021 (changes requested):
LGTM, but I need some time for test.
Thanks your help.

server/src/modes/script/componentInfo.ts
sapphi-red: I thought when there is a component like below, it will never emit `'foo-bar'` even if `@Emit` is declared.
```ts
@Component({
  emits: {
    foo: () => true
  }
})
export default class A extends Vue {
  @Emit()
  foo () { return true }

  @Emit()
  fooBar () { return true }
}
```
But yes it was not.

How should I handle it when it has a same event name in both `@Emit` and `emits`?
Just override `@Emit` one with `emits` one?

bissolli:
It doesn't give us type hints on what is coming from the event right? I've trying to find a good way to add a typed event but couldn't get anywhere close to a complete solution!

Found this [very nice package](https://github.com/andrewvasilchuk/vue-typed-emit), but I think it's not supported by Vetur - in the component where you use the event, it's not able to recognize the payload of the event!

Is there any good solution for that?

sapphi-red:
How about using `emits` option? See https://github.com/vuejs/rfcs/blob/master/active-rfcs/0030-emits-option.md#type-inference.

bissolli:
> How about using `emits` option? See https://github.com/vuejs/rfcs/blob/master/active-rfcs/0030-emits-option.md#type-inference.
> This PR obtains type data from this.

If I want to use it only for typing purpose I would need only to return true, right?
Any guesses on how long this PR will take to merge?

```
const Foo = defineComponent({
  emits: {
    submit: (payload: { email: string; password: string }) => true
  },

  methods: {
    onSubmit() {
      this.$emit('submit', {
        email: 'foo@bar.com',
        password: 123 // Type error!
      })

      this.$emit('non-declared-event') // Type error!
    }
  }
})
```

Also I am not very familiar with VSCode extensions, is there a way to use you PR to test it out?

yoyo930021:
> Also I am not very familiar with VSCode extensions, is there a way to use you PR to test it out?

https://github.com/vuejs/vetur/wiki/Developing-Language-Server-Features

server/src/modes/script/componentInfo.ts
sapphi-red: @yoyo930021 Do you have any opinions about the question above?

yoyo930021: There should be no problem.

yoyo930021 (approved):

server/src/modes/template/tagProviders/componentInfoTagProvider.ts
```diff
@@ -16,18 +16,21 @@ export function getComponentInfoTagProvider(childComponents: ChildComponent[]):
...
-    if (cc.info && cc.info.componentInfo.props) {
-      cc.info.componentInfo.props.forEach(p => {
-        props.push(genAttribute(`:${p.name}`, undefined, { kind: 'markdown', value: p.documentation || '' }));
+    const attributes: Attribute[] = [];
+    if (cc.info) {
+      cc.info.componentInfo.props?.forEach(p => {
+        attributes.push(genAttribute(`:${p.name}`, undefined, { kind: 'markdown', value: p.documentation || '' }));
+      });
+      cc.info.componentInfo.emits?.forEach(e => {
+        attributes.push(genAttribute(e.name, 'event', { kind: 'markdown', value: e.documentation || '' }));
```
yoyo930021, line 25: I find that I get a better experience without `@` char the beginning.
so I remove it.

yoyo930021:
LGTM, I will merge it when weekend.

The transforming template logic is changed in this PR.
@ktsn @octref can read it when you have time.

server/src/modes/script/componentInfo.ts
sapphi-red: @yoyo930021 
```ts
@Component({
  emits: {
    foo: (a: string) => true
  }
})
export default class A extends Vue {
  @Emit()
  foo () { return 0 }
}
```
This is an example which the type of listener will differ by overriding.
From `emits` the type becomes `(a: string) => void`, but from `@Emit` the type becomes `(a: number) => void`.

Overriding `@Emit` one with `emits` one will adopt `(a: string) => void`.

yoyo930021: Oh, it's a misunderstanding.
Your code is correct, so I don't change it.
You are right.

sapphi-red: I found vue-property-decorator does not support vue 3 for now and is going to deprecate `@Emit` (https://github.com/kaorun343/vue-property-decorator/issues/353).
I will change the test cases.

sapphi-red:
I fixed the test with class components with `@Options` since it was changed forVue 3.
Also added the test with property decorator with Vue 2.

PR #2565, opened 2020-12-13
Dont include used attributes in suggestions

sapphi-red:
This PR removes used attributes from suggestions except

- it is `class` , because `:class` and `class` can coexist (same with `style`)
  - https://eslint.vuejs.org/rules/no-duplicate-attributes.html
- it is a event listener, because it is able to listen to same event by adding modifiers

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+server/src/modes/template/test/completion.test.ts
--- a/server/src/modes/template/services/htmlCompletion.ts
+++ b/server/src/modes/template/services/htmlCompletion.ts
@@ -16,6 +16,7 @@ import { NULL_COMPLETION } from '../../nullMode';
 import { getModifierProvider, Modifier } from '../modifierProvider';
 import { toMarkupContent } from '../../../utils/strings';
 import { Priority } from '../tagProviders/common';
+import { kebabCase } from 'lodash';
 
 export function doComplete(
   document: TextDocument,
@@ -42,7 +43,6 @@ export function doComplete(
   const scanner = createScanner(text, node.start);
   let currentTag: string;
   let currentAttributeName = '';
-  let currentTagStartOffset: number;
 
   function getReplaceRange(replaceStart: number, replaceEnd: number = offset): Range {
     if (replaceStart > offset) {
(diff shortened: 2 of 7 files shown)


server/src/modes/template/services/htmlCompletion.ts
```diff
@@ -147,7 +148,11 @@ export function doComplete(
     return result;
   }
 
-  function collectAttributeNameSuggestions(nameStart: number, nameEnd: number = offset): CompletionList {
+  function collectAttributeNameSuggestions(
+    usedAttributes: string[],
```
merceyz, line 152: This would be much better suited as a `Set<string>` for performance

sapphi-red: Oh it should be, thank you!

server/src/modes/template/services/htmlCompletion.ts
```diff
@@ -392,3 +427,10 @@ function getWordEnd(s: string, offset: number, limit: number): number {
   }
   return offset;
 }
+
+function normalizeAttribute(attr: string): string {
+  // trim modifiers
+  attr = attr.replace(/\..+$/, '');
+
+  return attr.replace(/^(?:v-bind:|:)/, '').replace(/^v-on:/, '@');
```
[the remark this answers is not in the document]
yoyo930021: You can also consider this function.
https://github.com/vuejs/vetur/blob/b7ca3965f6a60b6413c1550788958b01de40d45e/server/src/modes/template/services/vuePropValidation.ts#L89

server/src/modes/template/services/htmlCompletion.ts
```diff
@@ -271,6 +286,23 @@ export function doComplete(
     return offset;
   }
 
+  function collectUsedAttributes(): Set<string> {
```
yoyo930021, line 289: You don't need this function.
`node.attributeNames` can get it.
Just like this https://github.com/vuejs/vetur/blob/b7ca3965f6a60b6413c1550788958b01de40d45e/server/src/modes/template/services/vuePropValidation.ts#L50.

.prettierignore
```diff
@@ -0,0 +1 @@
+server/src/modes/template/test/completion.test.ts
```
yoyo930021, line 1: Why do we need this?

yoyo930021 (changes requested):

.prettierignore
sapphi-red: Prettier breaks many lines by adding missing end tags and so on like this. (about 18 lines)
image

Should I put `<!-- prettier-ignore -->` comments on every line?

yoyo930021 (approved):

sapphi-red:
I rewrote it with Set. I have forgotten about the previous comment... https://github.com/vuejs/vetur/pull/2565#discussion_r542059708

PR #2609, opened 2021-01-01
Add stylelint-prettier support. Fix #712

### LINKED ISSUE #712: Prettier-Stylelint Support

otaq:
Based on https://github.com/vuejs/vetur/pull/1567.

### DIFF
--- a/package.json
+++ b/package.json
@@ -337,11 +337,13 @@
           "default": "prettier",
           "enum": [
             "none",
-            "prettier"
+            "prettier",
+            "stylelint-prettier"
           ],
           "enumDescriptions": [
             "disable formatting",
-            "css formatter using css parser from prettier"
+            "css formatter using css parser from prettier",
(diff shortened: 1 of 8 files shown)


yoyo930021:
Why do you close it?

otaq:
Just found out it's not working while writing integration test, will reopen when test is done.

PR #2627, opened 2021-01-07
Fix transition boolean props complete

sapphi-red:
To complete without `=""`, the type needed to be `"v"` instead of `"b"`.


yoyo930021:
I'm not sure this is a good change.
I am not sure about the definition of `type`.

It looks like changing here is also a possible direction.
https://github.com/vuejs/vetur/blob/5b89f1e03eca82a345b8eb4081a07d75ff0f7b89/server/src/modes/template/services/htmlCompletion.ts#L165

@octref What do you think?

sapphi-red:
`reversed` attribute of `ol` which is a boolean attribute[^1] has the type `v`, and `aria-atomic` attribute which should take `"true"` or `"false"`[^2] has the type `b`.
https://github.com/vuejs/vetur/blob/5b89f1e03eca82a345b8eb4081a07d75ff0f7b89/server/src/modes/template/tagProviders/htmlTags.ts#L155-L157
https://github.com/vuejs/vetur/blob/5b89f1e03eca82a345b8eb4081a07d75ff0f7b89/server/src/modes/template/tagProviders/htmlTags.ts#L528-L531

Vue's boolean attributes does allow a `value` contrary to HTML boolean attributes, but it also does not require a `value` contrary to ARIA's boolean attributes.

Changing that line to `(type !== 'v' || type !== 'b') && value.length` will effect ARIA's boolean attributes. So I think changing it to `v` is better.

[^1]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ol#attributes, https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes
[^2]: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions#advanced_live_regions

server/src/modes/template/tagProviders/vueTags.ts
```diff
@@ -54,10 +54,10 @@ const vueDirectives = [
 
 const transitionProps = [
   getAttribute('name', undefined, 'Used to automatically generate transition CSS class names. Default: "v"'),
-  getAttribute('appear', 'b', 'Whether to apply transition on initial render. Default: false'),
+  getAttribute('appear', 'v', 'Whether to apply transition on initial render. Default: false'),
   getAttribute(
     'css',
-    'b',
+    'v',
```
yoyo930021, line 60: This option default is true.
So we need `=""`.

server/src/modes/template/tagProviders/vueTags.ts
```diff
@@ -133,7 +133,7 @@ const vueTags = {
 const valueSets = {
   transMode: ['out-in', 'in-out'],
   transType: ['transition', 'animation'],
-  b: ['true', 'false']
+  v: ['true', 'false']
```
yoyo930021, line 136: It also need to keep `b`.

yoyo930021 (changes requested):

yoyo930021:
> `reversed` attribute of `ol` which is a boolean attribute[^1] has the type `v`, and `aria-atomic` attribute which should take `"true"` or `"false"`[^2] has the type `b`.
> https://github.com/vuejs/vetur/blob/5b89f1e03eca82a345b8eb4081a07d75ff0f7b89/server/src/modes/template/tagProviders/htmlTags.ts#L155-L157
> 
> 
> https://github.com/vuejs/vetur/blob/5b89f1e03eca82a345b8eb4081a07d75ff0f7b89/server/src/modes/template/tagProviders/htmlTags.ts#L528-L531
> 
> Vue's boolean attributes does allow a `value` contrary to HTML boolean attributes, but it also does not require a `value` contrary to ARIA's boolean attributes.
> 
> Changing that line to `(type !== 'v' || type !== 'b') && value.length` will effect ARIA's boolean attributes. So I think changing it to `v` is better.
> 
> [^1]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ol#attributes, https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes
> [^2]: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions#advanced_live_regions

I am not so familiar with this part.
I think some options may still be necessary.
@octref we need your help.

server/src/modes/template/tagProviders/vueTags.ts
sapphi-red: Oh yes it should be.

PR #2641, opened 2021-01-18
(dev) fix building in directory that has space in the path

jasonlyu123:
The server building would fail on a directory that has space in the path, for example: `c:/my projects/vetur`. Because the spawn of a child process of `tsc` uses the full path as a command like `c:/my projects/vetur/node_modules/.bin/tsc`.


yoyo930021 (approved):

yoyo930021:
LGTM, thanks for your help.

Issue #2642, opened 2021-01-21
improve template interpolation features performance when no changes to the file


jasonlyu123:
- [x] I have searched through existing issues

## Feature Request
Hi, First of all, thanks for the great works you guys put into it. The template interpolation feature really makes the developer experience way better and it's a good boost to productivity. But the performance is really not very good sometimes. I would like to help out and make it better. 

I'll focus on the performance of template interpolation features performance when no changes to the file in this issue. That is features like hover, completion resolve, go to definition, and find references. Especially the second request. I have done some investigation using my local clone version. Here's what I found out and think can make the performance better:

#### `vueInfoService.getInfo`
https://github.com/vuejs/vetur/blob/c26d51c342e495c27ab52dd39e22c7e68b6c6ab2/server/src/services/vueInfoService.ts#L105

This method is very demanding it often took more than 500ms on my machine. Besides the performance of this method, what strange to me is that the `vueFileInfo` map here seems like a kind of cache but it'll always get updated when `getInfo` is called. Furthermore, the stored info doesn't get removed when the document is closed or deleted in the file system, so it's a sort of memory leak. I think we could remove the map here.

https://github.com/vuejs/vetur/blob/85eecb3ada0a1d1bcdd916e2fa378f3fd708063d/server/src/modes/script/componentInfo.ts#L87

The most demanding part of  `vueInfoService.getInfo` seems to be this line. Analyzing the type afterward isn't that demanding. I notice typescript does cache the type checker result here. When typescript cache the type it runs significantly faster. **So the problem is mainly on why typescript won't use the cached version**. Once that's resolved. Most of the features that use this method might have a good boost in performance. 

Auto import, on the other hand, the implementation makes the script version mismatch on every resolve request. Thus typescript would always analyze it again. We probably need to find another way to improve it. I wonder if we can make the `mockDoc` [here](https://github.com/vuejs/vetur/blob/c26d51c342e495c27ab52dd39e22c7e68b6c6ab2/server/src/modes/plugins/autoImportSfcPlugin.ts#L57) different than the actual script doc.

#### `serviceHost.updateCurrentVirtualVueTextDocument`
https://github.com/vuejs/vetur/blob/c26d51c342e495c27ab52dd39e22c7e68b6c6ab2/server/src/services/typescriptService/serviceHost.ts#L194

[...]

PR #2645, opened 2021-01-23
improve performance of template interpolation features.

jasonlyu123:
#2642 

Added a test in the server to test if the type checker of typescript is reused because the performance test isn't easy. So I just test the main problem I found.


server/src/services/vueInfoService.ts
```diff
@@ -103,24 +103,20 @@ export interface MethodInfo {
 
 export class VueInfoService {
   private languageModes: LanguageModes;
-  private vueFileInfo: Map<string, VueFileInfo> = new Map();
```
yoyo930021, line 106: I want to keep this block code. It may be necessary to keep information about the whole project. We just don't use it yet.

In addition, we will soon abolish this mechanism.
This piece may be removed in the refactoring.

server/src/services/typescriptService/test/serviceHost.test.ts
```diff
@@ -0,0 +1,69 @@
...
+import { TextDocument } from 'vscode-languageserver-textdocument';
+import { URI } from 'vscode-uri';
+import { getDefaultVLSConfig } from '../../../config';
+import { getVueDocumentRegions } from '../../../embeddedSupport/embeddedSupport';
+import { LanguageModelCache } from '../../../embeddedSupport/languageModelCache';
+import { createEnvironmentService } from '../../EnvironmentService';
+import { getServiceHost } from '../serviceHost';
+
+suite('serviceHost', () => {
+  suite('interpolation performance', () => {
```
yoyo930021, line 13: I think the performance test may not be appropriate here.
Github action doesn't have stable operating resources.
This may cause our tests to be more unstable.

server/src/services/typescriptService/serviceHost.ts
```diff
@@ -192,13 +191,16 @@ export function getServiceHost(
...
     if (isVirtualVueTemplateFile(fileFsPath)) {
+      const oldDocVersion = localScriptRegionDocuments.get(fileFsPath)?.version;
       localScriptRegionDocuments.set(fileFsPath, doc);
       scriptFileNameSet.add(filePath);
       if (childComponents) {
         allChildComponentsInfo.set(filePath, childComponents);
       }
-      versions.set(fileFsPath, (versions.get(fileFsPath) || 0) + 1);
-      projectVersion++;
+      if (oldDocVersion !== doc.version) {
```
yoyo930021, line 200: Maybe it can include  line 195 code.

yoyo930021 (commented):
Sorry, I've been busy lately.
May not describe the problem well.
I think you've provided a good PR.

server/src/services/typescriptService/test/serviceHost.test.ts
jasonlyu123: We have some tests that use a typescript language service in another repo. We didn't encounter test timeout on GitHub action though. it usually only timeout on my slow office machine XD. 
With that said this test is more of a demonstration purpose. I guess there won't be many times people changing the logic here. I can remove the test.

yoyo930021:
LGTM, thank for your help.

PR #2666, opened 2021-02-01
Upgrade stylus-supremacy from 2.14.5 to 2.15.0

qnp:
When using Vetur on Vue SFC using stylus, we experienced a formatting bug caused by `stylus-supremacy` related to stylus that is colons were removed unwantedly for deep accessed properties. See https://github.com/ThisIsManta/stylus-supremacy/issues/79

The [2.15.0 release](https://github.com/ThisIsManta/stylus-supremacy/releases/tag/v2.15.0) includes the fix, so I suggest to upgrade it to make it available for the next Vetur release.

All tests passing ✓

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 0.32.1
+
+- Upgrade `stylus-supremacy` to 2.15.0
+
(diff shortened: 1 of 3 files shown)


yoyo930021:
LGTM, Thanks for your PR.

PR #2740, opened 2021-02-27
Improve Vue version detection

visualfanatic:
I installed Vue with `yarn add vue@3` and Yarn added `"vue": "3"` to `package.json` and then I configured Vetur to lint my code. For some reason I was getting `vue/no-multiple-template-root` error in a Vue 3 project, I removed Vue and then reinstalled it, but with a specific version and the error was gone. I changed the regexp inside `inferVueVersion` to handle such cases where only major version is saved in `package.json`.


yoyo930021 (approved):
Good catch, Thank you.

PR #2752, opened 2021-03-03
add --log-level option to vti diagnostics

### LINKED ISSUE #2723: Print only errors with VTI

## Feature Request

From https://github.com/vuejs/vetur/issues/1635#issuecomment-725332040.

To introduce VTI gradually, I'd like to check only errors.
The command will be like `vti diagnostics --error-only` I think.

Monchi:
Resolves #2723.

`vti diagnostics --log-level ERROR` to ignore warns and print only errors.

```bash
$ vti diagnostics --help
Usage: vti diagnostics [options] [workspace]

Print all diagnostics

Options:
  -l, --log-level <logLevel>  Log level to print (choices: "ERROR", "WARN", "INFO", "HINT", default: "WARN")
  -h, --help                  display help for command
```

### DIFF
--- a/vti/src/cli.ts
+++ b/vti/src/cli.ts
@@ -13,8 +13,10 @@ function getVersion(): string {
   program
     .command('diagnostics [workspace]')
     .description('Print all diagnostics')
-    .action(async workspace => {
-      await diagnostics(workspace);
+    .option('-e, --error-only', 'Print only errors', false)
+    .action(async (workspace, options) => {
+      console.log(options);
+      await diagnostics(workspace, options.errorOnly);
     });
 
   program.parse(process.argv);
--- a/vti/src/cli.ts
+++ b/vti/src/cli.ts
@@ -15,7 +15,6 @@ function getVersion(): string {
     .description('Print all diagnostics')
     .option('-e, --error-only', 'Print only errors', false)
     .action(async (workspace, options) => {
-      console.log(options);
       await diagnostics(workspace, options.errorOnly);
     });
 
--- a/vti/src/cli.ts
+++ b/vti/src/cli.ts
@@ -13,9 +13,18 @@ function getVersion(): string {
   program
     .command('diagnostics [workspace]')
     .description('Print all diagnostics')
-    .option('-e, --error-only', 'Print only errors', false)
+    .option('-l, --log-level <logLevel>', 'Log level to print', 'WARN')
     .action(async (workspace, options) => {
-      await diagnostics(workspace, options.errorOnly);
+      const logLevelOption: unknown = options.logLevel;
+      if (
+        logLevelOption !== 'ERROR' &&
+        logLevelOption !== 'WARN' &&
+        logLevelOption !== 'INFO' &&
+        logLevelOption !== 'HINT'
+      ) {
+        throw new Error(`Invalid log level: ${logLevelOption}`);
+      }
+      await diagnostics(workspace, logLevelOption);
     });
 
   program.parse(process.argv);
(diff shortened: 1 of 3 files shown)


vti/src/cli.ts
```diff
@@ -13,8 +13,10 @@ function getVersion(): string {
   program
     .command('diagnostics [workspace]')
     .description('Print all diagnostics')
-    .action(async workspace => {
-      await diagnostics(workspace);
+    .option('-e, --error-only', 'Print only errors', false)
+    .action(async (workspace, options) => {
+      console.log(options);
```
yoyo930021, line 18: we can remove it.

vti/src/cli.ts
```diff
@@ -13,8 +13,10 @@ function getVersion(): string {
   program
     .command('diagnostics [workspace]')
     .description('Print all diagnostics')
-    .action(async workspace => {
-      await diagnostics(workspace);
+    .option('-e, --error-only', 'Print only errors', false)
```
yoyo930021, line 16: The `DiagnosticSeverity` have four options.
Maybe we can make it a little more flexible.

yoyo930021 (changes requested):
Thank you.

vti/src/cli.ts
Monchi: @yoyo930021 I thought that the practical use case would be `--error-only` but flexibility is of course important. I've implemented `--log-level` option instead of `--error-only`.

BTW Currently `Information` and `Hint` severity would be printed as a warning. This is not in the scope of this PR and I would fix it in the other PR.

vti/src/cli.ts
```diff
@@ -13,8 +13,23 @@ function getVersion(): string {
...
-      await diagnostics(workspace);
+    .addOption(
+      new Option('-l, --log-level <logLevel>', 'Log level to print')
+        .default('WARN')
+        // logLevels is readonly array but .choices need read-write array (because of weak typing)
+        .choices((logLevels as unknown) as string[])
+    )
+    .action(async (workspace, options) => {
+      const logLevelOption: unknown = options.logLevel;
+      if (
```
yoyo930021, line 24: Why don't use `logLevels` from import?

vti/src/commands/diagnostics.ts
```diff
@@ -112,7 +115,20 @@ function range2Location(range: Range): SourceLocation {
   };
 }
 
-async function getDiagnostics(workspaceUri: URI) {
+function logLevel2Severity(logLevel: LogLevel): DiagnosticSeverity {
+  switch (logLevel) {
```
yoyo930021, line 119: Maybe using map of JS object replace `switch...case`?

yoyo930021 (changes requested):
LGTM

vti/src/cli.ts
Monchi: @yoyo930021 Because `logLevels` type is `readonly ["ERROR", "WARN", "INFO", "HINT"]`, we cannot call `logLevels.includes` with unknown value.

related issue: https://github.com/microsoft/TypeScript/issues/26255

vti/src/commands/diagnostics.ts
Monchi: sounds good, thanks!

vti/src/cli.ts
yoyo930021: We can use `as` keyword or `is` function to force it.
Repeated text will cause follow-up maintenance problems.

Monchi: @yoyo930021 ok. I've removed repeated text. e905cb2

yoyo930021 (approved):

Issue #2753, opened 2021-03-04
use vetur for files with extention other than `.vue`


lzl124631x:
I want to use vetur for files with extension `.wpy` instead of `.vue`. How to achieve this?

![image](https://user-images.githubusercontent.com/5824517/109924036-88e9b400-7c74-11eb-9fdd-76c2807388cb.png)

jasonlyu123:
https://code.visualstudio.com/docs/languages/identifiers

lzl124631x:
@jasonlyu123  Thanks. But I have that already I think?

![image](https://user-images.githubusercontent.com/5824517/110001325-21118880-7cc9-11eb-9f68-0f374d38f4d9.png)

Updated the screenshot. As you can see, Vetur is working but complaining my legit vue content.

jasonlyu123:
Oh sorry jumped to the conclusion too fast. There's some code using the `.vue` extension to check if a file is a vue file or ts/js files. Those need to be modified to allow a non-vue extension. 
for example:
https://github.com/vuejs/vetur/blob/7cf0b97fda4b4387ce2d7491f69522d5dbfce30b/server/src/services/typescriptService/serviceHost.ts#L301
The code here is used to tell typescript how to process the code. Maybe we can add a user config for other extensions.

On another note, maybe you can jump into it and try to implement it yourself. I am not a core maintainer but I think maybe this way it would have more chances to happens.

lzl124631x:
Do you mean adding a new config entry in [`vetur.config.js`](https://vuejs.github.io/vetur/reference/#example)? Something like the following? 

```
settings: {
  extensions: ["wpy"]
}
```

lzl124631x:
A rough PR for it https://github.com/vuejs/vetur/pull/2759/files

Haven't updated doc yet.

yoyo930021:
https://github.com/vuejs/vetur/issues/1616#issuecomment-584924652
https://github.com/vuejs/vetur/issues/2573#issuecomment-747184865

We have refused to do this before
But I'm not sure if there's a deeper reason.

lzl124631x:
[...]

PR #2759, opened 2021-03-05
add extentions to settings

lzl124631x:

### DIFF
--- a/server/src/config.ts
+++ b/server/src/config.ts
@@ -77,6 +77,7 @@ export interface VLSConfig {
     experimental: {
       templateInterpolationService: boolean;
     };
+    extensions: string[];
   };
 }
 
--- a/server/src/services/EnvironmentService.ts
+++ b/server/src/services/EnvironmentService.ts
@@ -11,6 +11,7 @@ export interface EnvironmentService {
   getVueVersion(): VueVersion;
   getSnippetFolder(): string;
   getGlobalComponentInfos(): BasicComponentInfo[];
+  isVueFile(path: string): boolean;
 }
 
 export function createEnvironmentService(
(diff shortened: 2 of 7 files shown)


server/src/config.ts
```diff
@@ -140,7 +141,8 @@ export function getDefaultVLSConfig(): VLSFullConfig {
       },
       experimental: {
         templateInterpolationService: false
-      }
+      },
+      extensions: []
```
yoyo930021, line 145: I think we can call `alias.extensions`

server/src/services/EnvironmentService.ts
```diff
@@ -35,6 +36,7 @@ export function createEnvironmentService(
     getPackagePath: () => packagePath,
     getVueVersion: () => inferVueVersion(packagePath),
     getSnippetFolder: () => snippetFolder,
-    getGlobalComponentInfos: () => globalComponentInfos
+    getGlobalComponentInfos: () => globalComponentInfos,
+    isVueFile: (path: string) => ['.vue', ...$config.vetur.extensions].some(ext => path.endsWith(ext))
```
yoyo930021, line 40: Maybe you need to prefix `.` when no dot input.

yoyo930021 (changes requested):
Can you add some documents?

lzl124631x:
> Can you add some documents?

Sure. But do we still need this change? This was originally for https://github.com/vuejs/vetur/issues/2753.

yoyo930021:
> > Can you add some documents?
> 
> Sure. But do we still need this change? This was originally for #2753.

In this comment https://github.com/vuejs/vetur/issues/2753#issuecomment-797184161 ,
Uni-app can use this PR. Of course if merge this PR, I can't disallow to use another extension.
But I think as long as it's stated in the document, it's fine.

PR #2802, opened 2021-03-23
Semantic tokens for typescript

### LINKED ISSUE #1904: FR: Support semantic color token customization

## Info

- Platform: all
- Vetur version: 0.24.0
- VS Code version: 1.45.0

## Problem

In my user settings I added the following:
```json
  "editor.semanticTokenColorCustomizations": {
    "[Default Dark+]": {
      "rules": {
        "*.readonly": "#9cdcfe",
        "enum": "#8f871f",
        "enumMember": "#bdb333"
      }
    }
  }
```

This works in  TypeScript files:
image

But does not seem to work in .vue files yet:
image

It would be nice if this is supported in vue files too.

### LINKED ISSUE #2434: Semantic highlight `.value` if using composition API
As referred to here: https://github.com/vuejs/rfcs/pull/222

Now Semantic Token is part of LSP (in the upcoming 3.16 release): https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/#textDocument_semanticTokens

We should adopt the latest LSP libraries and implement this feature to simplify composition API usage.

### LINKED ISSUE #2687: The syntax highlighting of lang="ts" is inconsistent with the syntax highlighting of the .ts file

## Info

- Platform: macOS
- Vetur version: 0.32.0
- VS Code version: 1.53.0
- VS Code theme: One Dark Pro
- VS Code formatter: Prettier

## Problem

image

This is an irrelevant question, but I have always been curious about what caused the inconsistency of the highlighting on both sides🤔 .
Prior to this, I have searched a lot of information and did not find the specific reason.

jasonlyu123:
Fixed #1904
Fixed #2687
Fixed #2434
Fixed #2687

### implentation 

Add semantic token feature for script region. Due to typescript's implementation. It's currently not available for js. Also because of that, this is currently not available in interpolation mode. But I still left some merging strategy in the project service for potential future implementation. 

A couple of things like to discuss. 

1. There's some new built-in vscode command in [1.53](https://code.visualstudio.com/updates/v1_53#_semantic-token-commands)
Wondering if I should update the integration test to use it. The advantage is it's more integrated than a language-server request. And  I don't have to copy the semantic token legend to the test script. A disadvantage is we might accidentally ship features not compatible with 1.52.  

2. Computed member is highlighted as a method. Should we need to override it?

### screenshot
Here is some screenshot comparison. The theme is GitHub dark.
Without semantic tokens
圖片

With semantic tokens
圖片

Without 
圖片

With
圖片

### DIFF
--- a/package.json
+++ b/package.json
@@ -516,6 +516,11 @@
           "default": true,
           "description": "Whether to automatic updating import path when rename or move a file"
         },
+        "vetur.languageFeatures.semanticTokens": {
+          "type": "boolean",
+          "default": true,
+          "description": "Whether to enable semantic highlighting. Currently only works for typescript"
+        },
         "vetur.trace.server": {
           "type": "string",
           "enum": [
--- a/package.json
+++ b/package.json
@@ -560,7 +560,14 @@
           "description": "Enable template interpolation service that offers hover / definition / references in Vue interpolations."
         }
       }
-    }
+    },
+    "semanticTokenScopes": [
+      {
+        "scopes": {
+          "property.refValue": ["entity.name.function"]
+        }
+      }
+    ]
   },
   "devDependencies": {
     "@rollup/plugin-commonjs": "^17.1.0",
--- a/server/src/services/projectService.ts
+++ b/server/src/services/projectService.ts
@@ -342,6 +342,12 @@ export async function createProjectService(
       return textDocumentEdit ?? [];
     },
     async onSemanticTokens(params: SemanticTokensParams | SemanticTokensRangeParams) {
+      if (!env.getConfig().vetur.languageFeatures.semanticTokens) {
+        return {
+          data: []
+        };
+      }
+
       const { textDocument } = params;
       const range = 'range' in params ? params.range : undefined;
       const doc = documentService.getDocument(textDocument.uri)!;
--- a/server/src/services/projectService.ts
+++ b/server/src/services/projectService.ts
@@ -22,6 +22,10 @@ import {
   FoldingRangeParams,
   Hover,
   Location,
+  SemanticTokens,
+  SemanticTokensBuilder,
+  SemanticTokensParams,
+  SemanticTokensRangeParams,
   SignatureHelp,
   SymbolInformation,
   TextDocumentEdit,
@@ -33,7 +37,7 @@ import { URI } from 'vscode-uri';
 import { LanguageId } from '../embeddedSupport/embeddedSupport';
 import { LanguageMode, LanguageModes } from '../embeddedSupport/languageModes';
 import { NULL_COMPLETION, NULL_HOVER, NULL_SIGNATURE } from '../modes/nullMode';
-import { DocumentContext, CodeActionData } from '../types';
+import { DocumentContext, CodeActionData, SemanticTokenData } from '../types';
 import { VCancellationToken } from '../utils/cancellationToken';
 import { getFileFsPath } from '../utils/paths';
 import { DependencyService } from './dependencyService';
@@ -60,6 +64,7 @@ export interface ProjectService {
   onCodeAction(params: CodeActionParams): Promise<CodeAction[]>;
   onCodeActionResolve(action: CodeAction): Promise<CodeAction>;
   onWillRenameFile(fileRename: FileRename): Promise<TextDocumentEdit[]>;
+  onSemanticTokens(params: SemanticTokensParams | SemanticTokensRangeParams): Promise<SemanticTokens | null>;
   doValidate(doc: TextDocument, cancellationToken?: VCancellationToken): Promise<Diagnostic[] | null>;
   dispose(): Promise<void>;
 }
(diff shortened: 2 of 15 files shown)


yoyo930021:
> There's some new built-in vscode command in 1.53
Wondering if I should update the integration test to use it. The advantage is it's more integrated than a language-server request. And I don't have to copy the semantic token legend to the test script. A disadvantage is we might accidentally ship features not compatible with 1.52.

My understanding is that you can use them.
Testing theoretically does not affect users.

> Computed member is highlighted as a method. Should we need to override it?

Do you know what it looks like in a typescript file?

jasonlyu123:
> My understanding is that you can use them.
Testing theoretically does not affect users.

Yes. It's just a very little chance that something doesn't work on an older version and we tested in a newer version so we didn't know. If you think it's ok I'll bump it.

>Do you know what it looks like in a typescript file? 
Should be the same as my current implementation. I haven't customized it. Because computed properties are method in the option object it has been highlighted as method not property.

yoyo930021:
> Yes. It's just a very little chance that something doesn't work on an older version and we tested in a newer version so we didn't know. If you think it's ok I'll bump it.

You can bump vscode version in the test. 👍

> Should be the same as my current implementation. I haven't customized it. Because computed properties are method in the option object it has been highlighted as method not property.

If same, I think we can keep it. Get feedback after the release.
Consider doing it together with this https://github.com/vuejs/vetur/issues/2434 .

jasonlyu123:
added #2434. Currently settle with a fallback TextMate scope of `entity.name.function`. If anyone can think of another better highlight scope, I can change it.

You can also customize the color follow [this doc](https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide#theming) after it goes live.

yoyo930021:
I will take a look on weekend.

server/src/services/vls.ts
```diff
@@ -384,6 +391,8 @@ export class VLS {
     this.lspConnection.onCodeAction(this.onCodeAction.bind(this));
     this.lspConnection.onCodeActionResolve(this.onCodeActionResolve.bind(this));
     this.lspConnection.workspace.onWillRenameFiles(this.onWillRenameFiles.bind(this));
+    this.lspConnection.onRequest(SemanticTokensRequest.type, this.onSemanticToken.bind(this));
+    this.lspConnection.onRequest(SemanticTokensRangeRequest.type, this.onSemanticToken.bind(this));
```
yoyo930021, line 395: Use `this.lspConnection.languages.semanticTokens.on` and `this.lspConnection.languages.semanticTokens.onRange` ?

server/src/services/projectService.ts
```diff
@@ -336,6 +341,38 @@ export async function createProjectService(
...
+      const range = 'range' in params ? params.range : undefined;
+      const doc = documentService.getDocument(textDocument.uri)!;
+      const modes = languageModes.getAllLanguageModeRangesInDocument(doc);
+      const data: SemanticTokenData[] = [];
+
+      for (const mode of modes) {
+        const tokenData = mode.mode.getSemanticTokens?.(doc, range);
+        // all or nothing
+        if (tokenData === null) {
+          return null;
```
yoyo930021, line 361: Why need this condition?
If some language broken, it probably disallows other languages.

package.json
```diff
@@ -555,7 +560,14 @@
...
       }
-    }
+    },
+    "semanticTokenScopes": [
+      {
+        "scopes": {
+          "property.refValue": ["entity.name.function"]
+        }
+      }
+    ]
```
yoyo930021, line 570: I also want to add this config.
```json
{
  "contributes": {
    "configurationDefaults": {
      "[vue]": {
        "editor.semanticTokenColorCustomizations": {
          "enabled": true,
          "rules": {
            "variable.ref": {
              "underline": true
            }
          }
        }
      }
    }
  }
}
```

yoyo930021 (changes requested):
LGTM
Sorry for the late reply.

server/src/services/vls.ts
jasonlyu123: The typing with that one doesn't allow a null result. But it's needed to disable full document semantic token for performance reason.

server/src/services/projectService.ts
jasonlyu123: I prefer it doesn't have a partial result for one language. But if you prefer the opposite I can remove it.

package.json
jasonlyu123: umm... vscode show warning about unknown editor configuration. Also, what is this`variable.ref` rule?

yoyo930021: Oh, I only want to make `Ref` underline.
Use `property.refValue` in this case.

server/src/services/vls.ts
jasonlyu123: I changed it to always return an empty array when a file is too big. i.e. completely disable semantic highlight on big files. So we don't need null anymore.

package.json
jasonlyu123: Doesn't seems to work. Maybe this config is not in effect here

yoyo930021: No problem. I can try it.

yoyo930021 (approved):
LGTM

Issue #2455, opened 2020-11-11
Allow passing files or single files to VTI 


soerenmartius:
- [X ] I have searched through existing issues

## Feature Request

At the moment it seems that the vti diagnostics command only accepts directories as an argument?
E.g., running vti diagnostics src/components/ will check all .vue inside the src/components directory. Passing a single file as an argument results in an ENOTDIR exception. It would be helpful to pass single files or a list of single files also use vti with pre-commit hooks also.

e.g.

```
vti diagnostics path/to/first/component.vue path/to/second/component.vue
```

yoyo930021:
we could do it with #2451.

soerenmartius:
> we could do it with #2451.

that'd be great!

soerenmartius:
> we could do it with #2451.
Hey @yoyo930021, do you plan on implementing this anytime soon?  I'd be more than happy to contribute as well!

yoyo930021:
> > we could do it with #2451.
> Hey @yoyo930021, do you plan on implementing this anytime soon?  I'd be more than happy to contribute as well!

You can do it.

PR #2825, opened 2021-03-26
add paths option to vti diagnostics

gregoirechauvet:
Resolves https://github.com/vuejs/vetur/issues/2455

`vti diagnostics --files src/App.vue src/Form.vue` to run VTI only on those two files

```
$ vti diagnostics --help
Usage: vti diagnostics [options] [workspace]

Print all diagnostics

Options:
  -l, --log-level <logLevel>  Log level to print (choices: "ERROR", "WARN", "INFO", "HINT", default: "WARN")
  -f, --files <files...>      Files to diagnose (default: null)
  -h, --help                  display help for command
```

### DIFF
--- a/vti/src/commands/diagnostics.ts
+++ b/vti/src/commands/diagnostics.ts
@@ -33,7 +33,7 @@ const logLevel2Severity = {
   HINT: DiagnosticSeverity.Hint
 };
 
-export async function diagnostics(workspace: string | null, logLevel: LogLevel) {
+export async function diagnostics(workspace: string | null, paths: string[], logLevel: LogLevel) {
   console.log('====================================');
   console.log('Getting Vetur diagnostics');
   let workspaceUri;
@@ -47,7 +47,7 @@ export async function diagnostics(workspace: string | null, logLevel: LogLevel) 
     workspaceUri = URI.file(process.cwd());
   }
 
-  const errCount = await getDiagnostics(workspaceUri, logLevel2Severity[logLevel]);
+  const errCount = await getDiagnostics(workspaceUri, paths, logLevel2Severity[logLevel]);
   console.log('====================================');
 
   if (errCount === 0) {
@@ -121,10 +121,28 @@ function range2Location(range: Range): SourceLocation {
   };
 }
 
-async function getDiagnostics(workspaceUri: URI, severity: DiagnosticSeverity) {
+async function getDiagnostics(workspaceUri: URI, paths: string[], severity: DiagnosticSeverity) {
   const clientConnection = await prepareClientConnection(workspaceUri);
 
-  const files = glob.sync('**/*.vue', { cwd: workspaceUri.fsPath, ignore: ['node_modules/**'] });
+  let files: string[];
+  if (paths.length === 0) {
+    files = glob.sync('**/*.vue', { cwd: workspaceUri.fsPath, ignore: ['node_modules/**'] });
+  } else {
+    // Could use `flatMap` once available:
+    const listOfPaths = paths.map(inputPath => {
+      const absPath = path.resolve(workspaceUri.fsPath, inputPath);
+
+      if (fs.lstatSync(absPath).isFile()) {
+        return [inputPath];
+      }
+
+      const directory = URI.file(absPath);
+      const directoryFiles = glob.sync('**/*.vue', { cwd: directory.fsPath, ignore: ['node_modules/**'] });
+      return directoryFiles.map(f => path.join(inputPath, f));
+    });
+
+    files = listOfPaths.reduce((acc: string[], paths) => [...acc, ...paths], []);
+  }
 
   if (files.length === 0) {
     console.log('No input files');
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Add `--log-level` option for `vti diagnostics` to configure log level to print. #2752.
 - 🙌 Semantic tokens for typescript and highlight `.value` if using composition API. Thanks to contribution from [@jasonlyu123](https://github.com/jasonlyu123). #2802 #1904 # 2434
+- Add paths option for `vti diagnostics` to diagnose only sub files or directories. #2455.
 
 ### 0.33.1 | 2021-03-07 | [VSIX](https://marketplace.visualstudio.com/_apis/public/gallery/publishers/octref/vsextensions/vetur/0.33.1/vspackage)
 
--- a/vti/src/cli.ts
+++ b/vti/src/cli.ts
@@ -15,7 +15,7 @@ function validateLogLevel(logLevelInput: unknown): logLevelInput is LogLevel {
   program.name('vti').description('Vetur Terminal Interface').version(getVersion());
 
   program
-    .command('diagnostics [workspace]')
+    .command('diagnostics [workspace] [paths...]')
     .description('Print all diagnostics')
     .addOption(
       new Option('-l, --log-level <logLevel>', 'Log level to print')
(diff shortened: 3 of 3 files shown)


rchl:
Maybe clarify how to pass multiple files? Separate paths with some delimiter or specify multiple `--files`. If the latter than "files" sounds kinda wrong from the user's perspective?

gregoirechauvet:
> because it's plural but accepts only one file?

Hum... no. It accepts an array of files. So you can do `--files a.vue b.vue c.vue`. Isn't it clear enough? Is there a way to make it clearer in the `--help` dialog?

Or you have another API in mind maybe?

rchl:
> Hum... no. It accepts an array of files. So you can do --files a.vue b.vue c.vue. 

Are you sure that even works? Typically in CLIs only one value is allowed for each option. So you would have to quote the value if you would want to include spaces in the value.

So I think that's not clear as it's against my expectations how CLIs work, at least.

> Or you prefer another API maybe?

I'll leave the final decision to the project owners.

rchl:
And since an optional `workspace` path is allowed at the end of the CLI, how would the parser know where one of the `--files` ends and workspace starts?

gregoirechauvet:
Yep, it works, it was running fine locally. For the workspace I was putting it before the `--files` option, like this:
```bash
vti diagnostics src/ --files a.vue b.vue c.vue
```

It's a valid behavior of commander: https://github.com/tj/commander.js#variadic-option

But you're right. I'm not sure of what to do, I need guidance here. Let me know how the API should look like :slightly_smiling_face:

gregoirechauvet:
> And since an optional workspace path is allowed at the end of the CLI, how would the parser know where one of the --files ends and workspace starts?

From commander's documentation, you can also use the special argument `--` to stop option processing:
```bash
vti diagnostics --files a.vue b.vue c.vue -- src/
```

rchl:
Ultimately it's up for module owners to decide but if it's supported by the CLI library used then I guess it's fine.

But I think it should be at least documented in https://github.com/vuejs/vetur/blob/master/docs/guide/vti.md, if not in the CLI itself.

rchl:
Also, it's maybe not clear what's the interaction when both the `--files` and `workspace` are specified.
And it's kinda weird that those are specified separately like that. Would it not be possible to just be able to specify file instead of the `workspace` argument, without have a separate `--files` argument?

I mean:

```
vti diagnostics -- a.vue b.vue c.vue
```

and then maybe one could also specify both directories and files together:

```
vti diagnostics -- src/ a.vue b.vue c.vue
```

gregoirechauvet:
> vti diagnostics -- a.vue b.vue c.vue

With this syntax, how to distinguish the workspace from the first file? With the `.vue` extension maybe?

> vti diagnostics -- src/ a.vue b.vue c.vue

And with this one, it might suggest that you can more than one workspace, no? Which isn't supported in VTI. Or that you can mix files and folders?

If this is the preferred way, I'm fine with it. Should I update the PR? Or should I wait for the project owners as you suggested once?

rchl:
Maybe don't listen to me but wait for owners of the code.

yoyo930021:
I think it would be nice to be consistent with the ESLint CLI.
https://eslint.org/docs/user-guide/command-line-interface#options

gregoirechauvet:
I've updated the PR to try to match ESLint syntax, as suggested. I haven't implemented the glob syntax, though.
```
$ vti diagnostics --help
Usage: vti diagnostics [options] [workspace] [paths...]

Print all diagnostics

Options:
  -l, --log-level <logLevel>  Log level to print (choices: "ERROR", "WARN", "INFO", "HINT", default: "WARN")
  -h, --help                  display help for command
```

From my understanding, VTI can't work without a workspace, right? Because it needs to find the `package.json` and such things.

Now after the workspace, you can add a list of files or directories. Their paths are relative to the workspace.
Then if you want to limit the diagnostics to a file, you also have to pass the workspace:
```
> vti diagnostics                                 // will run on the current directory as usual, the workspace is still optional
> vti diagnostics . src/App.vue                   // will run only on src/App.vue
> vti diagnostics . src/App.vue src/components/   // will run on src/App.vue and all .vue files inside src/components/
> vti diagnostics src/App.vue                     // will understand src/App.vue as the workspace, and fail
```

What do you think?

vti/src/commands/diagnostics.ts
```diff
@@ -121,10 +121,28 @@ function range2Location(range: Range): SourceLocation {
...
-  const files = glob.sync('**/*.vue', { cwd: workspaceUri.fsPath, ignore: ['node_modules/**'] });
+  let files: string[];
+  if (paths.length === 0) {
+    files = glob.sync('**/*.vue', { cwd: workspaceUri.fsPath, ignore: ['node_modules/**'] });
+  } else {
+    // Could use `flatMap` once available:
+    const listOfPaths = paths.map(inputPath => {
+      const absPath = path.resolve(workspaceUri.fsPath, inputPath);
+
+      if (fs.lstatSync(absPath).isFile()) {
```
gregoirechauvet, line 135: If the path doesn't exist, it'll fail with the following error:
```
> vti diagnostics . src/MissingComponent.vue

[...]
(node:1179198) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, lstat '/home/***/src/MissingComponent.vue'
    at Object.lstatSync (fs.js:1077:3)
    at /home/gregoire/Dev/vetur/vti/dist/cli.js:1:250515
    at Array.map (<anonymous>)
    at /home/gregoire/Dev/vetur/vti/dist/cli.js:1:250456
    at async Ji (/home/gregoire/Dev/vetur/vti/dist/cli.js:1:248898)
    at async a.<anonymous> (/home/gregoire/Dev/vetur/vti/dist/cli.js:1:252556)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:1179198) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:1179198) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```
Should I handle the error and display a clearer message maybe?

yoyo930021: This message should be clear enough.

yoyo930021:
> I've updated the PR to try to match ESLint syntax, as suggested. I haven't implemented the glob syntax, though.
> 
> ```
> $ vti diagnostics --help
> Usage: vti diagnostics [options] [workspace] [paths...]
> 
> Print all diagnostics
> 
> Options:
>   -l, --log-level <logLevel>  Log level to print (choices: "ERROR", "WARN", "INFO", "HINT", default: "WARN")
>   -h, --help                  display help for command
> ```
> 
> From my understanding, VTI can't work without a workspace, right? Because it needs to find the `package.json` and such things.
> 
> Now after the workspace, you can add a list of files and/or directories. Their paths are relative to the workspace.
> Then if you want to limit the diagnostics to a file, you also have to pass the workspace:
> 
> ```
> > vti diagnostics                                 // will run on the current directory as usual, the workspace is still optional
> > vti diagnostics . src/App.vue                   // will run only on src/App.vue
> > vti diagnostics . src/App.vue src/components/   // will run on src/App.vue and all .vue files inside src/components/
> > vti diagnostics src/App.vue                     // will understand src/App.vue as the workspace, and fail
> ```
> 
> What do you think?

It's well.
@octref Are you have some suggestions?

gregoirechauvet:
Hi. Is there anything I can do to help move this topic forward? :smile: 
Sorry if I'm being impatient.

yoyo930021:
LGTM, I merged it.

PR #2828, opened 2021-04-01
SugarSS Syntax Highlight

softwaredeveloptam:
- Adds SugarSS Syntax Highlight

Reference: [Syntax-SugarSS-SublimeText](https://github.com/aazcast/Syntax-SugarSS-SublimeText)


server/src/modes/style/index.ts
```diff
@@ -39,6 +39,15 @@ export function getPostCSSMode(
   return getStyleMode(env, 'postcss', languageService, documentRegions, dependencyService);
 }
 
+export function getSugarSSMode(
+  env: EnvironmentService,
+  documentRegions: LanguageModelCache<VueDocumentRegions>,
+  dependencyService: DependencyService
+): LanguageMode {
+  const languageService = getCSSLanguageService();
```
yoyo930021, line 47: If you use LSP, you need to add tests.
Replace `getSCSSLanguageService` to `getCSSLanguageService`?
https://github.com/MhMadHamster/vscode-postcss-language/blob/f2e7ff3be4d112fd22e4b5f43068a27c92652410/server/src/postcssServerMain.ts#L65

yoyo930021 (changes requested):

server/src/modes/style/index.ts
octref: Is using SCSS language service to support sugarss a good idea at all? As far as I know SCSS is not a superset of SugarSS. I think we should only support highlighting for SugarSS at most.

octref (commented):
https://github.com/hudochenkov/Syntax-highlighting-for-PostCSS/tree/master/Syntaxes Not sure if this is a better one, as it's more popular.

server/src/modes/style/index.ts
softwaredeveloptam: I spoke with @yoyo930021 a bit over Discord and it seems that SCSS/CSS language service aren't the best for SugarSS LSP. I've removed the LSP for now and will make another PR (or might have to create a `SugarSSLanguageService` then integrate it to future PR for Vetur)

softwaredeveloptam:
> https://github.com/hudochenkov/Syntax-highlighting-for-PostCSS/tree/master/Syntaxes Not sure if this is a better one, as it's more popular.

Both [hudochenkov's Syntax Highlighting for PostCSS](https://github.com/hudochenkov/Syntax-highlighting-for-PostCSS) and [aazcast's Syntax SugarSS SublimeText](https://github.com/aazcast/Syntax-SugarSS-SublimeText) have very similar grammar. [SugarSS](https://github.com/postcss/sugarss#text-editors) expresses that aazcast's SugarSS plugin is the go-to for SugarSS in Sublime 🙏. I hope this integration would be okay for Vetur 👍 . Very grateful for the chance to contribute to such a big resource to the Vue community 🙏 .

yoyo930021 (approved):
LGTM

PR #2879, opened 2021-04-21
Fix component data not shown in hover when template interpolation is on

### LINKED ISSUE #2878: Component data attribute description missing for dynamic attributes
- [x] I have read [FAQ](https://vuejs.github.io/vetur/guide/FAQ.html)

## Info

- Platform: macOS
- Vetur version: -
- VS Code version: -
- VLS version: 0.7.2

## Problem

The hover info shows the attribute description if the attribute is not dynamic (doesn't start with the `:`):

Screenshot 2021-04-21 at 15 38 24

But the description is nowhere to be found when the attribute is dynamic:

Screenshot 2021-04-21 at 15 38 03

## Reproducible Case

Repo: https://github.com/rchl/vetur-hover-component-data

1. Create component data like:

```
// tags.json
{
  "foo-tag": {
    "attributes": ["foo-attr"],
    "description": "A foo tag"
  }
}

// attributes.json
{
    "foo-attr": {
        "description": "Some property"
    }
}

// package.json
  "vetur": {
    "tags": "./tags.json",
    "attributes": "./attributes.json"
  }
```
2. Create a template:

```vue
<template>
  <div>
    <foo-tag foo-attr="v" /><foo-tag :foo-attr="'v'" />
  </div>
</template>
```

3. Hover the first and then the second `foo-attr`.

The first one shows the description from `attributes.json` while the second one doesn't.

rchl:
In cases where template interpolation provided some data for the hover
popup, it overrode the popup contents without letting component data
hover info to show.

Ask both services and merge the results together so that nothing is
lost.

Fixes #2878

### DIFF
--- a/server/src/modes/template/index.ts
+++ b/server/src/modes/template/index.ts
@@ -9,7 +9,6 @@ import { HTMLMode } from './htmlMode';
 import { VueInterpolationMode } from './interpolationMode';
 import { IServiceHost } from '../../services/typescriptService/serviceHost';
 import { HTMLDocument, parseHTMLDocument } from './parser/htmlParser';
-import { inferVueVersion } from '../../utils/vueVersion';
 import { DependencyService, RuntimeLibrary } from '../../services/dependencyService';
 import { VCancellationToken } from '../../utils/cancellationToken';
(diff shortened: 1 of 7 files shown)


yoyo930021 (approved):
LGTM

PR #2901, opened 2021-04-29
Fix attribute description not showing if its name matches html event

### LINKED ISSUE #2900: Component data doesn't work for attribute whose name matches HTML event
- [x] I have read [FAQ](https://vuejs.github.io/vetur/guide/FAQ.html)

## Info

- Platform: macOS
- VLS version: 0.7.2

## Problem

When a component has a property that matches the name of an existing HTML event then the component data does show up for that attribute when hovering it.

Hovering `error` attribute shows no popup in this case even though that custom attribute is defined in component data.

Screenshot 2021-04-29 at 23 13 36

## Reproducible Case

See `hover-error` branch on https://github.com/rchl/vetur-hover-component-data/tree/hover-error

rchl:
The code that matches attributes from all providers on hovering
attributes have found a match in the HTML tags provider (htmlTags.ts)
preventing the component data description from showing up.

Include '@' prefix in the collected HTML component data to not match
attributes without '@'. This makes sure that hovering "@click" attribute
matches the HTML provider and hovering "click" doesn't, for example.

This also improves (IMO) completion labels when adding event handlers
tags. Previously all labels for events from HTML provider completions
would not include the `@` prefix which wasn't obvious that it's an
event handler.

Fixes #2900

### DIFF
--- a/server/src/modes/template/tagProviders/htmlTags.ts
+++ b/server/src/modes/template/tagProviders/htmlTags.ts
@@ -880,7 +880,7 @@ export function getHTML5TagProvider(): IHTMLTagProvider {
     collectAttributes: (tag: string, collector: AttributeCollector) => {
       collectAttributesDefault(tag, collector, HTML_TAGS, globalAttributes);
       eventHandlers.forEach(handler => {
-        collector(handler, 'event');
+        collector('@' + handler, 'event');
       });
     },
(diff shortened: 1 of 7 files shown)


yoyo930021 (approved):
LGTM, Thank for your help.
Sorry for the late reply.

Issue #2908, opened 2021-04-30
Missing `.vue` extension in filename when using path/import autocompletion/intellisense


franciscolourenco:
- [x] I have searched through existing issues
- [x] I have read through [docs](https://vuejs.github.io/vetur)
- [x] I have read [FAQ](https://vuejs.github.io/vetur/guide/FAQ.html)
- [x] I have tried restarting VS Code or running `Vetur: Restart VLS`

## Info

- Platform: macOS
- Vetur version: v0.33.1
- VS Code version: 1.55.2

## Problem

When using the import path autocomplete/intellisense, `.vue` extension is not added automatically, however the extension is required for many of Vetur's functionalities. Is there an option to enable this?

![2021-04-30 14 12 46](https://user-images.githubusercontent.com/208149/116693418-2edc4500-a9be-11eb-869b-566b63e7e75c.gif)

## Reproducible Case

Use import path autocompletion on a fresh vue-cli project created with `vue create`

PR #2930, opened 2021-05-12
fix: 🐛 https://github.com/vuejs/vetur/issues/2908

IWANABETHATGUY:
fixed https://github.com/vuejs/vetur/issues/2908


IWANABETHATGUY:
cc @yoyo930021

IWANABETHATGUY:
`fileExtensionCompletion: false`

test1
`fileExtensionCompletion: true`

test2

rchl:
Does it have to be an option. I think that the extension is always required for vue files since that's how webpack knows how to parse it. It is not required or recommended for javascript/typescript files though so I think that:
 - there should be no setting
 - the logic should be specific to vue files

rchl:
Potentially the logic should be:
 - Skip extension for ts/js files
 - Add extension for any other type

IWANABETHATGUY:
> Potentially the logic should be:
> 
> * Skip extension for ts/js files
> * Add extension for any other type  

finished

yoyo930021 (changes requested):
The code is the right place to put it

https://github.com/vuejs/vetur/blob/4e54f396d9cafb1d7861ad8d17f8855fa635b07e/server/src/modes/script/javascript.ts#L252

yoyo930021 (approved):
LGTM, Thanks for your help.

PR #2943, opened 2021-05-17
Update project root shorthand comment

deisner:
The existing comment ("// shorthand for only root") confused me: I thought it meant there was only one project root, not that only the root is being specified.


yoyo930021:
LGTM, Thanks.
My English isn't well.

PR #3004, opened 2021-06-13
fix vetur.config.js links

fanvyr:


yoyo930021:
LGTM, Thanks.
We have same PRs.

PR #3009, opened 2021-06-16
Fix broken links in FAQ by removing /docs

DrKlonk:
There are some broken links in the FAQ page of the docs. This should fix them!

I tested it manually in the browser, not locally with the new urls.


yoyo930021:
LGTM, Thanks.
We have the same PRs.

PR #3015, opened 2021-06-20
Improve IntelliSense for component attributes/emits

triforcely:
This PR enables HTML mode for properties and emits since most of the logic was already implemented but failed for some reason. 

- Hover/Go to definition (F12) for components, properties and emits, with a correctly highlighted location in the source file.
- vetur.completion.attributeCasing configuration key for choosing how attributes should be formatted for auto-completion

### DIFF
--- a/server/src/embeddedSupport/languageModes.ts
+++ b/server/src/embeddedSupport/languageModes.ts
@@ -43,11 +43,13 @@ import { createAutoImportSfcPlugin } from '../modes/plugins/autoImportSfcPlugin'
 import { EnvironmentService } from '../services/EnvironmentService';
 import { FileRename } from 'vscode-languageserver';
 import { RefTokensService } from '../services/RefTokenService';
+import { DocumentService } from '../services/documentService';
 
 export interface VLSServices {
   dependencyService: DependencyService;
   infoService: VueInfoService;
   refTokensService: RefTokensService;
+  documentService: DocumentService;
 }
 
 export interface LanguageMode {
@@ -69,7 +71,7 @@ export interface LanguageMode {
   findDocumentHighlight?(document: TextDocument, position: Position): DocumentHighlight[];
   findDocumentSymbols?(document: TextDocument): SymbolInformation[];
   findDocumentLinks?(document: TextDocument, documentContext: DocumentContext): DocumentLink[];
-  findDefinition?(document: TextDocument, position: Position): Definition;
+  findDefinition?(document: TextDocument, position: Position, documentService: DocumentService): Definition;
   findReferences?(document: TextDocument, position: Position): Location[];
   format?(document: TextDocument, range: Range, options: FormattingOptions): TextEdit[];
   findDocumentColors?(document: TextDocument): ColorInformation[];
@@ -141,6 +143,7 @@ export class LanguageModes {
       this.documentRegions,
       autoImportSfcPlugin,
       services.dependencyService,
+      services.documentService,
       services.infoService
     );
 
--- a/server/src/modes/script/childComponents.ts
+++ b/server/src/modes/script/childComponents.ts
@@ -91,14 +91,25 @@ export function analyzeComponentsDefine(
           return;
         }
 
+        const definition = {
+          start: 0,
+          end: 0,
+          path: sourceFile.fileName
+        };
+
+        // If export node is equal to '{}' it means that the vue file doesn't have <script> component.
+        if (
+          defaultExportNode.kind !== tsModule.SyntaxKind.ObjectLiteralExpression ||
+          defaultExportNode.getText() !== '{}'
+        ) {
+          definition.start = defaultExportNode.getStart(sourceFile, true);
+          definition.end = defaultExportNode.getEnd();
+        }
+
         result.push({
           name: componentName,
           documentation: buildDocumentation(tsModule, definitionSymbol, checker),
-          definition: {
-            path: sourceFile.fileName,
-            start: defaultExportNode.getStart(sourceFile, true),
-            end: defaultExportNode.getEnd()
-          },
+          definition,
           defaultExportNode
         });
       }
--- a/server/src/modes/script/childComponents.ts
+++ b/server/src/modes/script/childComponents.ts
@@ -98,10 +98,7 @@ export function analyzeComponentsDefine(
         };
 
         // If export node is equal to '{}' it means that the vue file doesn't have <script> component.
-        if (
-          defaultExportNode.kind !== tsModule.SyntaxKind.ObjectLiteralExpression ||
-          defaultExportNode.getText() !== '{}'
-        ) {
+        if (!tsModule.isObjectLiteralExpression(defaultExportNode) || defaultExportNode.properties.length > 0) {
           definition.start = defaultExportNode.getStart(sourceFile, true);
           definition.end = defaultExportNode.getEnd();
         }
--- a/server/src/modes/template/parser/htmlParser.ts
+++ b/server/src/modes/template/parser/htmlParser.ts
@@ -77,9 +77,7 @@ export function parse(text: string): HTMLDocument {
   let endTagStart = -1;
   let pendingAttribute = '';
   let token = scanner.scan();
-  let startPos = -1;
-  let endPos = -1;
-  let attributes: { [name: string]: string } | undefined = {};
+  let attributes: { [k: string]: string } | undefined = {};
   while (token !== TokenType.EOS) {
     switch (token) {
       case TokenType.StartTagOpen:
@@ -139,8 +137,6 @@ export function parse(text: string): HTMLDocument {
         break;
       case TokenType.AttributeName:
         pendingAttribute = scanner.getTokenText();
-        startPos = scanner.getTokenOffset();
-        endPos = scanner.getTokenEnd();
         attributes = curr.attributes;
         if (!attributes) {
           curr.attributes = attributes = {};
@@ -149,8 +145,6 @@ export function parse(text: string): HTMLDocument {
         break;
       case TokenType.AttributeValue:
         const value = scanner.getTokenText();
-        startPos = scanner.getTokenOffset();
-        endPos = scanner.getTokenEnd();
         if (attributes && pendingAttribute) {
           attributes[pendingAttribute] = value;
           pendingAttribute = '';
--- a/server/src/modes/template/parser/htmlParser.ts
+++ b/server/src/modes/template/parser/htmlParser.ts
@@ -77,7 +77,9 @@ export function parse(text: string): HTMLDocument {
   let endTagStart = -1;
   let pendingAttribute = '';
   let token = scanner.scan();
-  let attributes: { [k: string]: string } | undefined = {};
+  let startPos = -1;
+  let endPos = -1;
+  let attributes: { [name: string]: string } | undefined = {};
   while (token !== TokenType.EOS) {
     switch (token) {
       case TokenType.StartTagOpen:
@@ -137,6 +139,8 @@ export function parse(text: string): HTMLDocument {
         break;
       case TokenType.AttributeName:
         pendingAttribute = scanner.getTokenText();
+        startPos = scanner.getTokenOffset();
+        endPos = scanner.getTokenEnd();
         attributes = curr.attributes;
         if (!attributes) {
           curr.attributes = attributes = {};
@@ -145,6 +149,8 @@ export function parse(text: string): HTMLDocument {
         break;
       case TokenType.AttributeValue:
         const value = scanner.getTokenText();
+        startPos = scanner.getTokenOffset();
+        endPos = scanner.getTokenEnd();
         if (attributes && pendingAttribute) {
           attributes[pendingAttribute] = value;
           pendingAttribute = '';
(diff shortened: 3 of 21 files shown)


server/src/modes/script/childComponents.ts
```diff
@@ -91,14 +91,22 @@ export function analyzeComponentsDefine(
...
         }
 
+        const definition = {
+          start: 0,
+          end: 0,
+          path: sourceFile.fileName
+        };
+
+        // If export node is equal to '{}' it means that the vue file doesn't have <script> component.
+        if (defaultExportNode.getText() !== '{}') {
```
rchl, line 101: I don't know much about walking the AST but I imagine getting the text of the whole script can be quite expensive. Is there no cheaper way to check check? Maybe just by checking the number of child nodes?

server/src/modes/template/services/htmlHover.ts
```diff
@@ -21,15 +23,18 @@ export function doHover(
...
+
+      function collect(resolvedTag: any, documentation: any) {
+        if (resolvedTag !== tag && kebabCase(resolvedTag) !== tag) {
           return;
         }
         hover = { contents: toMarkupContent(documentation), range };
-      });
+      }
+
+      provider.collectTags(collect);
```
rchl, line 37: This needs to be refactored since it crashes.
You can either created anonymous function and assign to variable or create function in outer scope and bind all needed variables to it.

rchl: And don't cheat with the types by using `any`. The `TagCollector` function is well defined in types.

yoyo930021 (changes requested):
I'm not sure this is the right direction.

Do we provide it with interpolation mode?

server/src/modes/template/parser/htmlParser.ts
```diff
@@ -77,7 +77,9 @@ export function parse(text: string): HTMLDocument {
   let endTagStart = -1;
   let pendingAttribute = '';
   let token = scanner.scan();
-  let attributes: { [k: string]: string } | undefined = {};
+  let startPos = -1;
+  let endPos = -1;
+  let attributes: { [name: string]: { value: string; start: number; end: number } } | undefined = {};
```
yoyo930021, line 82: Where are they using?

triforcely:
> I'm not sure this is the right direction.
> 
> Do we provide it with interpolation mode?

What do you mean? 

Interpolation mode is in an "experimental" state for years. We have logic for analyzing components, so it made sense to use it.

server/src/modes/template/services/htmlHover.ts
triforcely: Yeah, there should be a type. Previously, these collectors were just lambdas passed inline to "collectTags" function but I had some very strange results in the debugger. I decided to move these "collectors" outside and it fixed the issue but I forgot to set the types.

server/src/modes/script/childComponents.ts
triforcely: I'll see what can be done in this case.

server/src/modes/template/services/htmlHover.ts
rchl: It's not only about the types. The code actually crashes when using the compiled (`yarn compile`) code from the lib directory.

I'm not sure exactly why but you are creating a function definition inside a loop which JS is automatically hoisting so that's probably the reason.

rchl:
> Interpolation mode is in an "experimental" state for years. We have logic for analyzing components, so it made sense to use it.

Does it work with interpolation enabled though because with my brief testing pressing "go to definition" did nothing for me both for tags and attributes (for project-local components).

I run with those settings:
```
        "vetur.completion.autoImport": true,
        "vetur.experimental.templateInterpolationService": true,
        "vetur.useWorkspaceDependencies": true,
        "vetur.validation.templateProps": true,
```

yoyo930021:
You have misunderstood.
I was just wondering if I should put these things under `interpolationMode` file.

server/src/modes/template/parser/htmlParser.ts
triforcely: Hmm... I'll double-check in the evening but it looks like it ended up being unused.

server/src/modes/template/services/htmlHover.ts
triforcely: I'm suspecting that rollup is causing this issue, when I debugged "master" code I received pretty strange stuff as arguments, e.g. some native functions.

When you said that code crashes, does it crash after my fixes too?

triforcely:
> > Interpolation mode is in an "experimental" state for years. We have logic for analyzing components, so it made sense to use it.
> 
> Does it work with interpolation enabled though because with my brief testing pressing "go to definition" did nothing for me both for tags and attributes (for project-local components).
> 
> I run with those settings:
> 
> ```
>         "vetur.completion.autoImport": true,
>         "vetur.experimental.templateInterpolationService": true,
>         "vetur.useWorkspaceDependencies": true,
>         "vetur.validation.templateProps": true,
> ```

Are you running my branch or master? I just checked and it works fine on my branch, tests related to it pass too.

rchl:
> Are you running my branch or master? I just checked and it works fine on my branch, tests related to it pass too.

I've tested your branch of course.
Your branch crashes in the place I've mentioned in another issue so I had to manually fix that first.
Also the CI doesn't agree with you on the "passing all tests" part.

I'll just wait until you fix those issues and re-test then.

triforcely:
> > Are you running my branch or master? I just checked and it works fine on my branch, tests related to it pass too.
> 
> I've tested your branch of course.
> Your branch crashes in the place I've mentioned in another issue so I had to manually fix that first.
> Also the CI doesn't agree with you on the "passing all tests" part.
> 
> I'll just wait until you fix those issues and re-test then.

server/src/modes/script/childComponents.ts
triforcely: I'm afraid to do the check by the number of child nodes since it is hard to predict what kind of code users will input. I expanded the check to check first if node is ObjectLiteralExpression, so it can short-circuit earlier for most of the cases.

server/src/modes/script/childComponents.ts
```diff
@@ -91,14 +91,25 @@ export function analyzeComponentsDefine(
...
         }
 
+        const definition = {
+          start: 0,
+          end: 0,
+          path: sourceFile.fileName
+        };
+
+        // If export node is equal to '{}' it means that the vue file doesn't have <script> component.
+        if (
```
triforcely, line 101: @rchl @yoyo930021 This should short-circuit for majority of cases, without the need to call "getText".

triforcely:
> You have misunderstood.
> I was just wondering if I should put these things under `interpolationMode` file.
> 
> But I need more time to think.

Let me know if there are any changes that you would like regarding that.

BTW I'll fix tests later today.

server/src/modes/template/services/htmlHover.ts
rchl: Are you testing using code compiled for production (`vetur/server/bin/vls`)?

I can give you more details later but I'd just suggest not creating a local function declaration like that since it's not only bad practice but can also produce wrongly behaving code depending on if you use "var" or "let/const" in the loop. This is the relevant eslint rule that guards against doing that: https://eslint.org/docs/rules/no-inner-declarations

triforcely: ```yarn compile``` or ```yarn watch``` and then I use "all" launch target in vscode. All tests are green now. Let me know if the code in ```htmlHover.ts``` is ok after latest changes.

rchl: There is no crash anymore 👍

rchl:
Some notes after testing the latest code:
1. with those changes the "go to definition" on a tag referencing a local component goes to the beginning of the export in the script tag. On master it goes to the beginning of the vue file. I think the behavior on master might be preferred by many since people might want to "go to definition" with intent to modify the template. The behavior on this branch, even if looked from the internal logic might be "more correct", it might be less usable in practice
2. "go to definition" on a prop (attribute) places the cursor above the prop. For example here I've used "go to definition" on a `:show-title=true` prop and the cursor is on the line before that prop:
Screenshot 2021-06-21 at 22 26 04

triforcely:
> Some notes after testing the latest code:
> 
> 1. with those changes the "go to definition" on a tag referencing a local component goes to the beginning of the export in the script tag. On master it goes to the beginning of the vue file. I think the behavior on master might be preferred by many since people might want to "go to definition" with intent to modify the template. The behavior on this branch, even if looked from the internal logic might be "more correct", it might be less usable in practice
> 2. "go to definition" on a prop (attribute) places the cursor above the prop. For example here I've used "go to definition" on a `:show-title=true` prop and the cursor is on the line before that prop:
>    Screenshot 2021-06-21 at 22 26 04

Thanks for the feedback.
1. I usually like to navigate to components to take a look at the props and emits, to get an overview of what they expose. It works well in projects that have design systems and common components,
2. Yeah, I noticed that too. I'll traverse AST a little bit more to get a more specific location.

rchl:
> 1. I usually like to navigate to components to take a look at the props and emits, to get an overview of what they expose. It works well in projects that have design systems and common components.

I don't have strong opinion on that, I think. Will let @yoyo930021 decide.

yoyo930021:
> with those changes the "go to definition" on a tag referencing a local component goes to the beginning of the export in the script tag. On master it goes to the beginning of the vue file. I think the behavior on master might be preferred by many since people might want to "go to definition" with intent to modify the template. The behavior on this branch, even if looked from the internal logic might be "more correct", it might be less usable in practice
> I usually like to navigate to components to take a look at the props and emits, to get an overview of what they expose. It works well in projects that have design systems and common components.
> I don't have strong opinion on that, I think. Will let @yoyo930021 decide.

Maybe we can add a option?
I will choose `<script>` when only one.

> "go to definition" on a prop (attribute) places the cursor above the prop. For example here I've used "go to definition" on a :show-title=true prop and the cursor is on the line before that prop:
> Yeah, I noticed that too. I'll traverse AST a little bit more to get a more specific location.

This problem may not need to traverse AST.
Try to replace `ast.pos` or `ast.end` to `ast.getStart` or `ast.getEnd` functions.

server/src/modes/script/childComponents.ts
yoyo930021: Maybe  `defaultExportNode.properties.length === 0`

triforcely:
> > with those changes the "go to definition" on a tag referencing a local component goes to the beginning of the export in the script tag. On master it goes to the beginning of the vue file. I think the behavior on master might be preferred by many since people might want to "go to definition" with intent to modify the template. The behavior on this branch, even if looked from the internal logic might be "more correct", it might be less usable in practice
> > I usually like to navigate to components to take a look at the props and emits, to get an overview of what they expose. It works well in projects that have design systems and common components.
> > I don't have strong opinion on that, I think. Will let @yoyo930021 decide.
> 
> Maybe we can add a option?
> I will choose `<script>` when only one.
> 
> > "go to definition" on a prop (attribute) places the cursor above the prop. For example here I've used "go to definition" on a :show-title=true prop and the cursor is on the line before that prop:
> > Yeah, I noticed that too. I'll traverse AST a little bit more to get a more specific location.
> 
> This problem may not need to traverse AST.
> Try to replace `ast.pos` or `ast.end` to `ast.getStart` or `ast.getEnd` functions.
> 
> https://ts-ast-viewer.com/#code/KYDwDg9gTgLgBAE2AMwIYFcA28DeAoOQuMKCMAZwC458i64BLcgUQDtUAjTYa2+-mAE8wPOAGUYUBqwDmBfnAC+AGnn8mAESaduvNQsJCR1CVNn66i-VcVA

Thank you, I'm not that proficient in TS compiler API and didn't know that there was a difference between field and getter function. After changing it to getStart and getEnd it feels much more polished now :)

server/src/modes/script/childComponents.ts
triforcely: @yoyo930021 I replaced it with your suggestion.

server/src/embeddedSupport/languageModes.ts
```diff
@@ -69,7 +71,7 @@ export interface LanguageMode {
   findDocumentHighlight?(document: TextDocument, position: Position): DocumentHighlight[];
   findDocumentSymbols?(document: TextDocument): SymbolInformation[];
   findDocumentLinks?(document: TextDocument, documentContext: DocumentContext): DocumentLink[];
-  findDefinition?(document: TextDocument, position: Position): Definition;
+  findDefinition?(document: TextDocument, position: Position, documentService: DocumentService): Definition;
```
yoyo930021, line 74: We don't add this argument.
We can send directly for `documentService` to `LanguageMode`.

yoyo930021 (changes requested):
Sorry, I'm late.
Only have single probably problem.
LGTM for others

carvemerson:
`+1`

triforcely:
Hey @yoyo930021 . I apologize for such a long delay. I'll update branch, try to fix mentioned issues and hopefully we can get this merged.

triforcely:
I'm sorry but I'm not going to be able to fix all of these. Too much time has passed since my fork was created and it is not a matter of simply solving merge conflicts but a lot of things have changed in the project area (files moved, removed, refactored, etc). I'm not using vue that much anymore and simply cannot force myself to spend a few evenings after work to make this PR happen.

Sorry to everybody who counted on this one but you can still use general ideas and parsing logic from my branch if you ever wanted to port them to the latest master.

PR #3222, opened 2021-11-01
fix(interpolation/v-for): allow optional index parameter for numeric iteration

thebanjomatic:
The `v-for` directive when iterating over a number provides both the 1-based numeric value, and the 0-based index to the iteration function. Vetur currently shows this as an error when you are using the `experimental.templateInterpolationService` option and attempt to use the index.

Existing behavior:
image
```
No overload matches this call.
  The last overload gave the following error.
    Argument of type 'number' is not assignable to parameter of type 'object'.Vetur(2769)
```

This change adds the missing second parameter to the iteration helper for the `num: number` overload, and existing usages with a single parameter should continue to work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### Un release
+
+- 🙌 Add support to interpolation service for index parameter with number iteration. Thanks to contribution from [@thebanjomatic](https://github.com/thebanjomatic). #3222
+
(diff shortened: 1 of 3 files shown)


thebanjomatic:
@yoyo930021 sorry to ping you, but I wasn't sure if you've seen this PR yet.

yoyo930021 (approved):
LGTM

PR #3348, opened 2022-02-15
docs: add BalmUI for supported framework

elf-mouse:
Hey there,

BalmUI recently released Vetur support for the Vue 3 integration in our v9.42.0 release and the Vue 2 integration in our v8.49.0 release.

We would love it if we could get listed on the "Supported Frameworks" section.

Thanks!


yoyo930021:
LGTM, thanks.

PR #3353, opened 2022-02-18
Add support for checking `default` and `required` state of props set by variables (`Identifier`).

lordeleto:
This PR allows componentInfo to inspect the `default` and `required` properties of a prop that is set by a variable, i.e. a canned prop declaration.

Example:
CommonProps.ts
```
export const LabelProp = {
  type: String,
  required: true,
};

export const IsValidProp = {
  type: Boolean,
  default: false,
};
```
SampleComponent.vue
```
import Vue from 'vue';
import { LabelProp, IsValidProp } from './CommonProps.ts';

export default Vue.extend({
  name: 'SampleComponent',
  props: {
    label: LabelProp,
    valid: IsValidProp
  }
});
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Un release
 
 - 🙌 Add support to interpolation service for index parameter with number iteration. Thanks to contribution from [@thebanjomatic](https://github.com/thebanjomatic). #3222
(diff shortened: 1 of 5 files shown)


yoyo930021:
LGTM, thanks.

